PR #4, opened 2020-11-02
Improve developer experience

leio10:
This PR prepares the BB to be used and tested from Decidim and addresses some issues we had:

*  Improves the developer experience: adds support for VSCode remote containers, create dev keys for dev decidim instances (we will add those keys to Decidim too), improves spec support and uses a different port than Decidim by default.

* Applies some design changes for the whole solution: uses headers instead of arguments for client authentication, uses JWK format for public keys instead of PEM and add unique ids for clients and elections, to allow to refer them with the same on every part of the system.

* Adds missing tests for the create_election endpoint.

* Adds support for the `iat` keys on the JWT messages.

* Changes the way authentication for clients is done.


agustibr (approved):
👏 These are really great improvements!

app/commands/create_election.rb
```diff
@@ -47,93 +50,84 @@ def build_election
...
+                              "Invalid signature"
+                            elsif decode_error.present?
+                              decode_error
+                            elsif election.voting_scheme_class.blank?
+                              "A valid Voting Scheme must be specified"
+                            elsif title.blank?
+                              "Missing title"
+                            elsif start_date.after?(end_date)
+                              "Starting date cannot be after the end date"
+                            elsif start_date.before?(Time.current + 2 * 60 * 60)
```
slickepinne, line 111: Can we maybe use `2.hours.from_now` instead of `(Time.current + 2 * 60 * 60)` ? I think, it's easier to understand.

slickepinne (approved):
Great! 👍 I just left a small comment. 😄

app/commands/create_election.rb
leio10: Oh, I didn't see that one! Thanks! ❤️

beagleknight (approved):
LGTM! 👍 I just added a small suggestion that can be addressed later. I would also add to the PR that this also refactors the `CreateElection` command to use the new concern `LogEntryCommand` 👍

app/commands/concerns/log_entry_command.rb
```diff
@@ -0,0 +1,45 @@
...
+    @log_entry = LogEntry.new(
+      signed_data: signed_data,
+      log_type: type,
+      client: client
+    )
+  end
+
+  def log_entry_validations
+    if decoded_data.blank?
+      "Invalid signature"
```
slickepinne, line 18: Oh, one more comment. 😄 Will we use Internationalization? If so, we might want to add it before we end up with lots of locale strings.

leio10: yes, we will address that in the future, probably adding an error code to the message.

app/commands/create_election.rb
```diff
@@ -47,93 +50,84 @@ def build_election
...
-  def chained_hash
-    Digest::SHA256.hexdigest(signed_data)
+      if number_elected.blank?
+        "There must be specified the number of answers to be selected"
+      elsif ballot_selections.blank? || ballot_selections.size <= 1
+        "There must be at least 2 answers for each question"
+      elsif number_elected.to_i > ballot_selections.size
+        "The number of possible answers cannot be greater than the number of answers offered"
+      end
+    end.compact.first
```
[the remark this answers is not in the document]
leio10: yes, and having the client gem in this repo will make it much more easier :+1:

beagleknight (approved):

slickepinne (approved):

leio10:
@decidim/core in this PR we commited some improvements and refactors needed for the key ceremony PR to work with smaller PRs, and we will implement the user stories on a new PR, depending on this one. So, there is no need of approval from product. Can you review it please? Thanks!

agustibr (approved):

leio10:
(this PR was closed to change the branch name)

PR #6, opened 2020-11-09
Improve developer experience

leio10:
This PR prepares the BB to be used and tested from Decidim and addresses some issues we had:

*  Improves the developer experience: adds support for VSCode remote containers, create dev keys for dev decidim instances (we will add those keys to Decidim too), improves spec support and uses a different port than Decidim by default.

* Applies some design changes for the whole solution: uses headers instead of arguments for client authentication, uses JWK format for public keys instead of PEM and add unique ids for clients and elections, to allow to refer them with the same on every part of the system.

* Adds missing tests for the create_election endpoint.

* Adds support for the `iat` keys on the JWT messages.

* Changes the way authentication for clients is done.


leio10:
@decidim/core in this PR we commited some improvements and refactors needed for the key ceremony PR to work with smaller PRs, and we will implement the user stories on a new PR, depending on this one. So, there is no need of approval from product. Can you review it please? Thanks!

agustibr (approved):

tramuntanal (approved):
Great refactoring! :smiley:

PR #8, opened 2020-11-11
Move rails application to its own folder

beagleknight:
**Important ⚠️ :** Even the PR looks like a lot of changes I am just moving things around so keep that in mind while reviewing it 👍 

This PR moves all the Ruby on Rails application code to its own folder `decidim-bulletin-board-app`. I also changed the Github workflow to run in the specific folder and only when there is a change on that folder.
As an extra bonus ✨ I added some cache so the build is a bit faster.

The idea is to have the application in a separate folder because we are gonna add two more projects to the repository: 
- See https://github.com/decidim/decidim-bulletin-board/pull/9 to know more about the new ruby gem that will be used to interact with the bulletin board application from server code.
- See https://github.com/decidim/decidim-bulletin-board/pull/10 to know more about the new npm package that will be used to interact with the bulletin board application from client code.


README.md
```diff
@@ -1,39 +1,5 @@
...
-```
-bundle exec rubocop
-```
-
-
-* How to run the test suite
-```
-bundle exec rspec
-```
+- **decidim-bulletin-board-app**: A Ruby on Rails application that contains the Bulletin Board service.
```
leio10, line 5: maybe we can link the app readme from here, what do you think?

leio10 (approved):
👏 👏 Great job, step by step the pieces begin to fall into place! 💪

README.md
beagleknight: Yeah! I think it will be a great addition 😄

beagleknight: @leio10 I added the link! It will work when it's merged to `master`. Can you double check please? Thanks!

leio10 (approved):

agustibr (approved):

slickepinne (approved):

PR #15, opened 2020-11-18
Integrate electionguard wrapper

leio10:
🎩 **What? Why?**

Adds the electionguard voting scheme to the Bulletin Board, using the Python wrapper for the Bulletin Board implemented in [codegram/decidim-electionguard](https://github.com/codegram/decidim-electionguard) (in the future this repo will be under the decidim account). 

It adds to the documentation and the VS container the dependencies for the electionguard (python and gmpy2 required libraries) and for running python code from a ruby app.

It also refactors the processing of the messages and the validations, to simplify that process as much as possible.

To be able to process messages, the electionguard wrappers needs to have different types of messages for each phase (key_ceremony or tally). To support this in the BB application, this PR converts the type field to a `message_id`, which includes all the context information needed about the message without having to decode the signed message. This also removes the need to add the trustee identification in the headers to identify their messages.

Finally, it adds the new voting scheme, acting as a proxy between the app and the python wrapper.

📌 **Related Issues**
Related to decidim/decidim#6722


leio10:
@decidim/core This PR doesn't have a related issue, so it doesn't need to be tested by @decidim/product and it's ready to be reviewed. Thanks!

slickepinne (approved):
Very nice! I added a small note only. Nothing that would block the merge. 😊

beagleknight (approved):
Looks great! 🤩

agustibr (approved):
👏👏 Congrats, great work!!

tramuntanal (approved):
LGTM!
Although I can't add much value regarding business logic

PR #17, opened 2020-11-19
chore(app): allow AJAX requests from different domains

beagleknight:
This adds the `rack-cors` gem to allow AJAX requests form a different domain. By default it only allows `localhost:3000` since it is the domain used for the `decidim` application in development.

In production environments a suitable domain must be provided using the `CORS_ORIGIN_ALLOWED` environment variable.

### DIFF
--- a/decidim-bulletin-board-app/config/initializers/cors.rb
+++ b/decidim-bulletin-board-app/config/initializers/cors.rb
@@ -1,0 +1,10 @@
+# frozen_string_literal: true
+
+return if Rails.application.config.settings.cors_origin_allowed.blank?
+
+Rails.application.config.middleware.insert_before 0, Rack::Cors do
+  allow do
+    origins "*"
+    resource "*", headers: :any, methods: [:get, :post, :patch, :put]
+  end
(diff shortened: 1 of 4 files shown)


decidim-bulletin-board-app/config/initializers/cors.rb
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+
+return if Rails.application.config.settings.cors_origin_allowed.blank?
+
+Rails.application.config.middleware.insert_before 0, Rack::Cors do
+  allow do
+    origins "*"
```
agustibr, line 7: Should this be replaced by the environment variable?

```suggestion
    origins Rails.application.config.settings.cors_origin_allowed
```

beagleknight: Oh wow... I totally forgot haha. Good catch @agustibr !

leio10 (approved):
👏 👍

agustibr (approved):
🚀

PR #18, opened 2020-11-19
feat: bulletin board ruby gem

slickepinne:
This PR adds a new bulletin board ruby gem to be used by Decidim and the Bulletin Board app.


decidim-bulletin_board-ruby/spec/decidim/bulletin_board_spec.rb
```diff
@@ -1,11 +1,7 @@
 # frozen_string_literal: true
```
beagleknight, line 1: We can remove this test now 😄

decidim-bulletin_board-ruby/spec/fixtures/bb_schema.json
```diff
@@ -0,0 +1,1806 @@
+{
```
beagleknight, line 1: I'm not sure if we need to check out this file in the repository yet 🤔  Let's leave it for now, but we should think about that in the future

beagleknight (approved):
Awesome ✨

agustibr (approved):
👏

PR #20, opened 2020-11-19
feat: enhance bulletin board graphql schema to support the client

beagleknight:
This adds a new field to the `Query` object to get an election by its `unique_id`. This is used to get all the log entries for a specific `election`.

Then I added a new service called `LogEntryNotifier` that will notify all the subscribers to the `election_log_entry_added` event and used it in the `ProcessKeyCeremonyStep` command.


decidim-bulletin-board-app/app/graphql/types/query_type.rb
```diff
@@ -29,6 +29,17 @@ def elections
       Election.all
     end
 
+    field :election,
+          Types::ElectionType, 
+          null: true,
+          description: "Returns an election given its unique_id" do
+      argument :unique_id, String, required: true
```
leio10, line 36: We are using the `unique_id` as the `id` for the whole graphql interface. So, I would prefer to keep calling it `id` in this endpoint. But I understand that this could be confusing (as the database `id` is not used and the `unique_id` is exposed as `id`). Do you think we should change this line to match the rest of the app or change it everywhere to make it more simpler (use `unique_id` instead of `id` everywhere)?

decidim-bulletin-board-app/app/commands/process_key_ceremony_step.rb
```diff
@@ -32,6 +32,7 @@ def call
 
       election.log_entries << log_entry
       log_entry.save!
+      LogEntryNotifier.new(log_entry).notify_subscribers
```
leio10, line 35: ❤️ this!
Should we do this as the last step, to avoid notifying users after rollbacking the transaction. I know it is not big deal, but I like to let these kind of steps for the end of the transaction.

decidim-bulletin-board-app/app/commands/process_key_ceremony_step.rb
```diff
@@ -69,12 +70,16 @@ def process_message
...
 
-    LogEntry.create!(
+    log_entry = LogEntry.create!(
       election: election,
       signed_data: BulletinBoard.sign(response_message),
       log_type: :key_ceremony,
       bulletin_board: true
     )
+
+    LogEntryNotifier.new(log_entry).notify_subscribers
```
leio10, line 80: again, maybe we can store this log entry as `response_log_entry` and wait after saving all the models to notify the users

leio10 (changes requested):
Great work! I'd only address the `unique_id`/`id` issue before merging this. 💪

decidim-bulletin-board-app/app/graphql/types/query_type.rb
agustibr: I'm in favor to use `unique_id` everywhere, as @leio10 commented just to don't generate confusion, as it doesn't refer to the db `id` but to a custom identifier.

leio10: Well, I have doubts about that. It makes complete sense inside the app, as we don't use the same name for different things. But from the API consumer point of view it's a bit strange, all ids are supposed to be unique, why don't we just call it `id`?
ElectionGuard uses something called `object_id` to store these slug based identifiers. It would be nice to use that, but all Ruby objects already have an `object_id`, so we can't use that name.
Also, something similar could happen with message identifiers. I didn't use `unique_id` because I'm not sure if they are going to be unique, but my idea was to use them as the `id` field. The problem could arise if we have two messages with the same id, so maybe we should revert this to use the autonumeric field.

beagleknight: Honestly... I agree with you both 😂 . I understand that maybe this is giving to much details to the API consumer so calling it just `id` is ok for me even if internally we are using the `unique_id`. TBH I used `id` first and then I changed it but now I am having second thoughts. Maybe using another identifier different from `id` and `unique_id` can help?

decidim-bulletin-board-app/app/commands/process_key_ceremony_step.rb
beagleknight: Oh totally! I forgot about the transaction. I'll move it outside of it!

decidim-bulletin-board-app/app/commands/process_key_ceremony_step.rb
beagleknight: Yeah, I'll do that! 👍

beagleknight:
Hey @leio10 ! I applied the changes at https://github.com/decidim/decidim-bulletin-board/pull/20/commits/426b9ab4a131626a15c212d69dd4003e959b1cc8 Can you take a look to them? Thanks!

leio10 (approved):
Great! I don't have a clear idea about the `unique_id` naming issue, but I think we should move on and maybe address it in the future.

PR #21, opened 2020-12-01
feat: add key ceremony events observable

beagleknight:
This adds a simple `events` observable to the `KeyCeremony` class. As a user of the library now I can do this:

```js
const keyCeremony = new KeyCeremony(...);

keyCeremony.events.subscribe(event => {
  // do something with the event
});
```

At the moment there are just two events supported: `MESSAGE_RECEIVED` and `MESSAGE_PROCESSED`. These two events can be exported from the key ceremony module as well:

```js
const { MESSAGE_RECEIVED, MESSAGE_PROCESSED } = decidimBulletinBoard.KeyCeremony;
```

This can be useful to update the UI accordingly in case you are expecting some messages to be processed.

### DIFF
--- a/decidim-bulletin_board-js/package-lock.json
+++ b/decidim-bulletin_board-js/package-lock.json
@@ -1,5 +1,5 @@
 {
-  "name": "decidim-bulletin-board",
+  "name": "decidim-bulletin_board",
   "version": "1.0.0",
   "lockfileVersion": 1,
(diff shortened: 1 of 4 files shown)


leio10 (approved):
🔝 👍

slickepinne (approved):
Very nice!  :muscle:

PR #23, opened 2020-12-01
Save election with pending messages

leio10:
After adding the identifiers to the messages we can get the election identifier without having to decode the signed message. So, its better to assign the election to the pending messages before enqueueing them to be processed. This PR implements this.


beagleknight (approved):
Nice! 👍

agustibr (approved):
Great, this is improving! 😀

PR #24, opened 2020-12-02
Activate election BB endpoint

agustibr:
This PR adds the `OpenBallotBox` mutation and command, so when the election is on a `:ready` status the Authority can request the activation of the election (open the ballot box) and change the election status to `:vote`.

Related to:
- https://github.com/decidim/decidim/issues/6632


leio10 (approved):
Awesome! 💃 👏

decidim-bulletin_board-app/app/graphql/mutations/open_ballot_box_mutation.rb
```diff
@@ -0,0 +1,30 @@
...
+
+    field :election, Types::ElectionType, null: true
+    field :error, String, null: true
+
+    def resolve(message_id:, signed_data:)
+      authority = find_authority(message_id)
+
+      return { error: "Authority not found" } unless authority
+
+      result = { error: "There was an error opening the election's ballot box." }
```
beagleknight, line 16: This is never used, right?

decidim-bulletin_board-app/spec/commands/open_ballot_box_spec.rb
```diff
@@ -0,0 +1,84 @@
...
+    let(:trustee) { election.trustees.first }
+    let(:message_id) { "#{election.unique_id}.open_ballot_box+t.#{trustee.unique_id}" }
+    let(:message) { build(:open_ballot_box_message, message_id: message_id, election: election, trustee: trustee) }
+
+    it "broadcast invalid" do
+      expect { subject }.to broadcast(:invalid)
+    end
+  end
+
+  # TODO: test race conditions
```
beagleknight, line 83: Did you forget something or are you going to implement this in a future PR? I don't like leaving TODO messages in the code but that's just a personal preference 😄

beagleknight (approved):
Good job! 👍  I just left some comments but everything looks good.

decidim-bulletin_board-app/app/graphql/mutations/open_ballot_box_mutation.rb
leio10: Well, that's inherited from my mutations, and its related to our beloved Rectify. At the beginning I was returning the result directly from the `on(...)` block. But that fails sometimes, so I created this local variable. As it could be possible that none of the `on` blocks are executed, this default value would be easier to debug than receiving an empty response.

decidim-bulletin_board-app/spec/commands/open_ballot_box_spec.rb
leio10: This is also inherited from the key ceremony part. In this case is not needed, as there will be no race conditions problems for this part. I'll work on the TODO as soon as possible 😅

decidim-bulletin_board-app/app/commands/open_ballot_box.rb
```diff
@@ -0,0 +1,46 @@
...
+  # Returns nothing.
+  def call
+    return broadcast(:invalid, error) unless
+      valid_log_entry?("open_ballot_box")
+
+    election.with_lock do
+      return broadcast(:invalid, error) unless
+        valid_step?(election.ready?) &&
+        valid_client?(authority.authority?) &&
+        valid_author?(message_identifier.from_authority?)
```
leio10, line 32: I think we should add a `&& process_message` call here, to allow voting schemes to be informed about this messages reception (also for the close_ballot_box message). We don't need to modify the dummy scheme and it won't do anything with the message, but the EG wrapper will use it.

decidim-bulletin_board-app/spec/commands/open_ballot_box_spec.rb
agustibr: Good catch, I'll remove the `TODO` comment 😁

decidim-bulletin_board-app/app/commands/open_ballot_box.rb
agustibr: I was in doubt if the `process_message` was needed here, in any case I'll add it.

leio10 (approved):
LGTM!

slickepinne (approved):
Great!!! 👏

PR #26, opened 2020-12-07
chore: deprecate logType for log entries

beagleknight:
This deprecated the `logType` field since it has been replaced by the `messageId`.


leio10 (approved):
Great! I thought I had already done this 😅

PR #22, opened 2020-12-01
Add vote endpoint

leio10:
Add the vote endpoint to the Bulletin Board, to receive and verify the ballots casted by the voters.

Related to decidim/decidim#6625

This PR also includes some improvements and refactors done before to simplify its implementation:
* Save the election reference with pending messages (#23)
* Improve the base classes for mutations 163ccd1 
* Improve the dummy voting scheme response generation ebe52d4

PR #27, opened 2020-12-07
fix: setup election now needs a message id

beagleknight:
This PR started fixing a simple 🐛 but we ended adding more features to make the development a bit easier.

When the GraphQL schema changes our clients need to be aware, so I created to handy tasks to keep them up to date:

```
$ bundle exec rake schema:generate
$ npm run schema:generate
```

These two commands will generate a dump schema file in the ruby and js clients. If you forget to run these the GitHub Workflow will fail so a developer cannot forget about these.

For the Ruby part the dump schema file will be used to run the tests, so we can detect breaking changes immediately. For the JS part the schema generation uses our actual operation files in `decidim-bulletin_board-js` so it will also fail if something is not correct.

Finally and the main reason to do this PR was to update both the "setup election" and the "process key ceremony step" mutation usage across the clients using the `messageId` argument introduced in #22


agustibr (approved):
👍

decidim-bulletin_board-ruby/lib/decidim/bulletin_board/client.rb
```diff
@@ -34,6 +34,11 @@ def configured?
         private_key && server && api_key
       end
 
+      def setup_election(election_data)
+        message_id = "#{election_data[:election_id]}.create_election+a.#{authority_slug}"
+        Decidim::BulletinBoard::CreateElection.call(election_data, message_id)
```
leio10, line 39: The `election_data` has a `:message_id` entry, so maybe we could get it directly from there to be more DRY.
Actually, we have discussed about this with @slickepinne, and we reached the conclusion that we should move the JSON building code to the gem (in the future). 
After doing that, I don't have a clear idea about where is the best place to calculate the `message_id`. But I'd say that the responsible should be the `CreateElection` class, maybe using a concern to reuse the common part of all message identifiers. What do you think?

leio10 (approved):
:+1:

slickepinne (approved):
👏 Nice!!

decidim-bulletin_board-ruby/lib/decidim/bulletin_board/client.rb
beagleknight: Yeah! We had the same discussion 😄 We should definetly move it to the gem as well. About the `message_id` generation I'm not sure and I had the same problem with the JS client. I put it in the `KeyCeremony` class for the moment but doesn't feel right.

I think that it will be much clear when we start implementing the EG wrapper and dealing with different messages maybe. For the Ruby part maybe it's better to have it in the `CreateElection` class as you said. We can do that when we refactor the whole `election_data` stuff 👍

PR #28, opened 2020-12-07
Validate client type and message author in commands

agustibr:
This PR validates that the command is called by the correct Client and also ensures that the message author is the expected, otherwise it broadcast an `:invalid`.


beagleknight (approved):
LGTM! 👍

slickepinne (approved):
Nice!! 😄

leio10 (approved):
Great job! 👏 🔝

PR #30, opened 2020-12-08
feat: cast vote from the ruby client

beagleknight:
This PR includes a single method on the Bulletin Board client: `cast_vote`. This is a part of #25 and it will be used after the vote has been encrypted in the browser.

The idea is to use this to publish the encrypted vote to the Election log. I added a few dependencies to use forms and commands like we are doing in other places.

From `decidim` we will need to call this method like these:
```ruby
# assuming that the encrypted vote has been received from the user performing the
# operation via an HTTP request.

bulletin_board_client.cast_vote(
  { election_id: "test.1" },
  { voter_id: "voter.1" }
  encrypted_vote
)
```


decidim-bulletin_board-ruby/lib/decidim/bulletin_board/client.rb
```diff
@@ -34,13 +34,18 @@ def configured?
...
+
       def setup_election(election_data)
         message_id = "#{election_data[:election_id]}.create_election+a.#{authority_slug}"
         Decidim::BulletinBoard::CreateElection.call(election_data, message_id)
       end
 
-      def cast_vote
-        cast_vote = Decidim::BulletinBoard::Voter::CastVote.new
+      def cast_vote(election_data, voter_data, encrypted_vote)
+        form = Decidim::BulletinBoard::Voter::VoteForm.new(self, election_data, voter_data, encrypted_vote)
```
leio10, line 47: Did you think about exposing the form to the app and receiving it as an argument? That way you would be able to validate the data before calling the method, returning error messages and would be a good approach for the `create_election` method. What do you think?

leio10 (approved):
Nice!! 🎉

decidim-bulletin_board-ruby/lib/decidim/bulletin_board/client.rb
beagleknight: I really don't know 😄! I don't like `decidim` knowing too much about implementation details. I mean with the current implementation decidim just needs to know this:

```ruby
begin
  bulletin_board_client.cast_vote(...)
rescue
  # something bad happens (i.e. the provided data is wrong)
end
```

That means it doesn't need to know about forms or commands or any internal stuff, and we can change the gem implementation when we want without changing the public API.

But I also agree that using just `Hash` could be very error prone, so I am not 100% sure about the previous statement. I agree that we should refactor the `create_election` method to hide the `election_data` implementation details, but I am not sure how to do it at the moment.

When we do that refactor we can discuss the different approaches if you want. Sounds good? 😄

slickepinne (approved):
Great! I'll use that approach for the `create_election` then. 👍

PR #32, opened 2020-12-09
Close ballot box BB endpoint

agustibr:
This PR adds the `CloseBallotBox` mutation and command to the Bulletin Board, so when the election is on `:vote` status the Authority can request the closing of the election (close the ballot box) and change the election status to `:tally`.

Related to:

- https://github.com/decidim/decidim/issues/6634


decidim-bulletin_board-app/app/commands/close_ballot_box.rb
```diff
@@ -0,0 +1,47 @@
...
+  # - :invalid if the received data wasn't valid.
+  #
+  # Returns nothing.
+  def call
+    return broadcast(:invalid, error) unless
+      valid_log_entry?("close_ballot_box")
+
+    election.with_lock do
+      return broadcast(:invalid, error) unless
+        valid_step?(election.vote?) &&
```
leio10, line 30: This reminds me that, at the moment, we are not adding date validations on this endpoint. We should have this in mind in the future, to double check that is the desired behavior.

leio10 (approved):
Great job! 👍 👏

decidim-bulletin_board-app/app/commands/close_ballot_box.rb
agustibr: Oh yeah, in previous iterations of the _open ballot box_ it was checked in the process message.
I could check here that the date/time is valid.

beagleknight (approved):
LGTM! ✨

leio10 (approved):
:+1:

PR #33, opened 2020-12-09
feat: client gets election status

slickepinne:
With this PR, decidim can receive the status for a specific `election` given the `election.id`. The `election_id` gets created on the client.

Example `decidim`:

```        
        def election_status
          bulletin_board.get_status(election.id)
          .... #do something with status
        end
```


leio10 (approved):
Lovely! 👏

beagleknight (approved):
Nice! ✨ Just one thing that I wanted to mention is now we are raising `StandardError` and that's very nice. We'll refactor this later with our Exceptions 👍

agustibr (approved):
Great!

slickepinne:
> Nice! ✨ Just one thing that I wanted to mention is now we are raising `StandardError` and that's not very nice. We'll refactor this later with our Exceptions 👍

Yeah, I just used `StandardError` as Graphlient suggested to use that one instead of `ServerError`.

PR #36, opened 2020-12-10
fix: sign data was not using the correct private key to sign

beagleknight:
This fixes the `sign_data` method to use the correct private key. I also added the missing test to catch the 🐛


agustibr (approved):
🚀

slickepinne (approved):
👍

leio10 (approved):

PR #37, opened 2020-12-10
New messages format

leio10:
While connecting the JS client with the BB we detected some inconsistencies in the messages format that needs to be addressed. So, we have defined a new message format, isolating the keys used by the BB and the information handled by the wrappers, like this:

```json
{
  "iat": 123456,
  "message_id": "decidim-barcelona.1.key_ceremony.trustee_election_keys+t.trustee-1",
  "content": "here goes the wrapper information"
}
```

This way, is possible to calculate the SHA for the content generated by the wrapper before completing and signing the message.

This PR apply the new format in the bulletin board application and also to the key ceremony part of the JS client.

### DIFF
--- a/decidim-bulletin_board-app/Gemfile.lock
+++ b/decidim-bulletin_board-app/Gemfile.lock
@@ -293,6 +293,7 @@ GEM
 
 PLATFORMS
   ruby
+  x86_64-linux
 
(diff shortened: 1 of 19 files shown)


beagleknight (approved):
Great! ✨

PR #29, opened 2020-12-08
Add support for vote verification

leio10:
This PR adds the content hash field to log entries and allow clients to retrieve a log entry by its content hash. This feature will be used to implement the vote verification described in decidim/decidim#5975. 
It also adds the `messageId` field to the PendingMessage GraphQL type and a test for the `pendingMessage` query.

PR #39, opened 2020-12-11
Add a JS client method to get a LogEntry by hash

leio10:
This PR adds a method to the JS client to use the endpoint implemented in #29 to retrieve a log entry using the hash of the log entry content.

### DIFF
--- a/decidim-bulletin_board-js/src/client/operations/get_log_entry.gql
+++ b/decidim-bulletin_board-js/src/client/operations/get_log_entry.gql
@@ -1,0 +1,9 @@
+query GetElectionLogEntries($electionUniqueId: String!, $contentHash: String!) {
+  logEntry(electionUniqueId: $electionUniqueId, contentHash: $contentHash) {
+    logEntry {
+      messageId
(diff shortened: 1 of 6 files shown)


decidim-bulletin_board-js/src/client/operations/get_log_entry.gql
```diff
@@ -0,0 +1,9 @@
+query GetElectionLogEntries($electionUniqueId: String!, $contentHash: String!) {
+  logEntry(electionUniqueId: $electionUniqueId, contentHash: $contentHash) {
+    logEntry {
```
beagleknight, line 3: Is this correct? Looks a bit odd that you have two `logEntry`. I would expect the query to be like this:
```graphql
query {
  logEntry(...) {
    messageId
    signedData
  }
}
```

Otherwise the shape of your response will be `data.logEntry.logEntry.messageId`

beagleknight (approved):
LGTM! Just added a comment in the query because I think our schema might be wrong.

PR #40, opened 2020-12-11
Simplify key ceremony in the JS client

leio10:
This PR modifies the key ceremony in the JS client to simplify the flow of calls. It allows to `run` several times, stopping the message processing when a message requires to save the wrapper state or when the process is done. In case of having to save the wrapper state, the response message will be saved to be sent on the next `run` call.
It also address some issues with naming and code organization.


slickepinne (approved):
Looks great! Can you add the move of the `create_election` command to the PR description? It's not really related to the `key_ceremony`. 😄

PR #42, opened 2020-12-15
Refactor bulletin board JS client

slickepinne:
This PR exports the `MessageIdentifier` to be used in the `key_cermeony` process on Decidim. It also adds the `backup` method of the `TrusteeWrapper` to `KeyCeremony` to be used directly on Decidim calling the `key_ceremony`.

### DIFF
--- a/decidim-bulletin_board-js/src/key-ceremony/key-ceremony.js
+++ b/decidim-bulletin_board-js/src/key-ceremony/key-ceremony.js
@@ -61,6 +61,14 @@ export class KeyCeremony {
   }
 
   /**
+   * Returns the backup function of the TrusteeWrapper
+   * @returns {String}
+   */
+  backup() {
+    return this.currentTrustee.wrapper.backup()
+  }
+
+  /**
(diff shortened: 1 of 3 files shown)


decidim-bulletin_board-js/src/key-ceremony/key-ceremony.js
```diff
@@ -60,6 +60,14 @@ export class KeyCeremony {
...
   }
 
+  /**
+   * Returns the backup function of the TrusteeWrapper
+   * @returns {String}
+   */
+  backup() {
+    return this.currentTrustee.wrapper.backup()
+  }
+
```
leio10, line 70: Should we add the restore method too, or can we wait until needing It?

leio10 (approved):
Great work!! 👏👏👏

decidim-bulletin_board-js/src/key-ceremony/key-ceremony.js
slickepinne: I would wait until we actually need it. 😄

PR #43, opened 2020-12-16
Improve wrapper state handling

leio10:
This PR allows the key ceremony process to continue from a previous process, without sending duplicated messages to the Bulletin Board.


slickepinne (approved):
Really nice! ❤️ 👍

PR #44, opened 2020-12-17
Open and close ballot box from authority

agustibr:
This PR adds the methods for opening and closing the ballot box to the Bulletin Board client.


slickepinne (approved):
Perfect! 😃 We shouldn't forget to handle the `Graphlient` errors maybe a bit more specific.

PR #45, opened 2020-12-18
Use the BulletinBoard gem in the BB app.

leio10:
This PR removes duplicated code, using the `JwkUtils` and `MessageIdentifier` classes from the gem. It also fixes a warning with GraphQL subscriptions type definition.


slickepinne (approved):
Great work! 💪

PR #46, opened 2020-12-18
Refactor gem commands to improve consistency

leio10:
The client commands were starting to be very different from each others. This PR fixes that creating a base class and normalizing the `client` methods.


slickepinne (approved):
Amazing work!!! ❤️ 
I just have one note: Can you update the gem version and release it with this PR? 🙌

leio10:
> Amazing work!!! ❤️
> I just have one note: Can you update the gem version and release it with this PR? 🙌

Yeah! My plan is to that after merging the PR with the open and close commands.

PR #52, opened 2020-12-20
Fix graphql schema file location

leio10:
When using the gem in the app, it doesn't load the schema right. This PR fixes that, using absolute paths for the json file path.


slickepinne (approved):
👍

PR #54, opened 2020-12-21
Fix notifier errors

leio10:
This PR tries to fix the errors raised when triggering events inside transaction, as sometimes the receiver code can't access to the rows created in the transaction until this is commited.


agustibr (approved):
👍

slickepinne (approved):
🙏 Wonderful!!!

PR #55, opened 2020-12-21
Add verify vote and get the vote process working

leio10:
This PR adds the verify vote method to the voter client. 

Also, it allows the contentHash to be null on the GraphQL types, to support returning the messages without a `content` key. 

Finally, it updates the changelog, includes the prettier extension in the dev containers and updates the changelog with some missing changes.


agustibr (approved):

slickepinne (approved):
Great! Really awesome! 😍

PR #57, opened 2020-12-23
Configuration improvements to deploy to the staging environment

leio10:
This PR adds some changes to be able to test better the app and to deploy it to the staging env. We should review this parameters in the future.


slickepinne (approved):
💪

PR #60, opened 2020-12-28
Replace GraphQL subscriptions with polling

leio10:
Using GraphQL subscriptions for the key ceremony raises issues with race conditions and make tests harder to reproduce on CI servers. This PR uses polling for that, that makes the process more predictable and it's easier to mock in the Decidim Elections module.

This PR also adds a production build command, that should be used to generate the JS files used for the production app.


decidim-bulletin_board-js/src/key-ceremony/key-ceremony.test.js
```diff
@@ -1,23 +1,27 @@
...
-    getElectionLogEntries: jest.fn(() => Promise.resolve([])),
-    subscribeToElectionLogEntriesUpdates: jest.fn((_args, fn) =>
-      electionLogEntriesUpdates.pipe(tap(fn)).subscribe()
-    ),
-    processKeyCeremonyStep: jest.fn(),
+    getElectionLogEntries: jest.fn(() => {
+      const result = [];
+      for (let i = 0; i < electionLogEntriesUpdates.length; i++) {
+        result.push(electionLogEntriesUpdates.shift());
+      }
```
beagleknight, line 19: If you just want to copy the values from one array to the other you can do something like this:
```js
const result = [...electionLogEntriesUpdates];
```

decidim-bulletin_board-js/src/test-utils.js
```diff
@@ -0,0 +1,5 @@
+function buildMessageId(typeSubtype) {
+  return `decidim-barcelona.1.${typeSubtype}+a.decidim-barcelona`;
+}
+
+export { buildMessageId };
```
beagleknight, line 5: You can also use:

```js
export function buildMessageId() { ... }
```

decidim-bulletin_board-js/src/trustee/__mocks__/trustee.js
```diff
@@ -1,6 +1,13 @@
+import { buildMessageId } from "../../test-utils";
+
 export class Trustee {
   processLogEntry({ messageId, signedData }) {
-    switch (messageId) {
+    const typeSubtype = messageId
+      .split("+")[0]
+      .split(".", 4)
+      .slice(2, 4)
+      .join(".");
```
beagleknight, line 9: I am a bit worried about having these implementation details here. Maybe we can create a function in the `test-utils` file?

decidim-bulletin_board-js/webpack.prod.js
```diff
@@ -0,0 +1,11 @@
...
+const { CleanWebpackPlugin } = require("clean-webpack-plugin");
+const common = require("./webpack.common.js");
+
+module.exports = merge(common, {
+  mode: "production",
+  plugins: [new CleanWebpackPlugin()],
+  output: {
+    filename: "[name].prod.js",
+  },
+});
```
beagleknight, line 11: Neat 🤩

beagleknight (approved):
Good job! I really think this will simplify many things and for this use case subscriptions are not that important. I added some comments, mostly suggestions. For the tests I think that we should consider adding some kind of factories library or so.

decidim-bulletin_board-js/src/key-ceremony/key-ceremony.js
```diff
@@ -37,46 +38,33 @@ export class KeyCeremony {
...
-        messageIdentifier.author.type === TRUSTEE_TYPE &&
-        messageIdentifier.author.id === this.currentTrustee.id
-      ) {
-        this.trusteeSentMessageIds.push(message.messageId);
-      }
-    });
+  async getLogEntries() {
+    const { id: electionUniqueId } = this.electionContext;
+
+    const lastLogEntry = this.electionLogEntries.slice(-1)[0];
```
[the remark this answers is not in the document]
leio10: nope, I don't want to modify the array. Anyway, I think that this version is not clearer than `this.electionLogEntries[this.electionLogEntries.length - 1]`, which is faster, so I will change it.

slickepinne (approved):
Very nice! 👍

PR #61, opened 2020-12-29
Rails engine and decidim-bulletin_board.js as asset

agustibr:
This PR changes the gem to be a Rails Engine, also builds the js in the gem assets directory.


decidim-bulletin_board-js/webpack.common.js
```diff
@@ -1,18 +1,16 @@
...
 
-module.exports = (env) => ({
-  mode: env.development ? "development" : "production",
+module.exports = {
   entry: "./src/index.js",
   output: {
-    path: path.resolve(__dirname, "dist"),
-    filename: "[name].js",
+    path: path.resolve(__dirname, "..", "decidim-bulletin_board-ruby", "app", "assets","javascripts", "decidim", "bulletin_board"),
+    filename: "decidim-bulletin_board-s.js",
```
beagleknight: Is the `-s` a typo?

beagleknight (approved):
It looks great! 🤩 I am just worried about forgetting to run the build command before releasing the gem but maybe we can just add a CI step on the `decidim-bulletin_board-js` folder or something like that.

decidim-bulletin_board-js/webpack.common.js
agustibr: Yes, it's fixed now 😁

leio10 (approved):
🤩 Great work!!

slickepinne (approved):
Awesome!! 🎉

PR #65, opened 2021-01-15
Simplify tests and improve their security and performance

leio10:
This PR refactors the tests system to make them easier to create. Instead of creating everything for each test, now we use seeds to create the authority and the three trustees used for development also for the tests. That way, we can use the well known private keys for signing the messages created, instead of having to create new private keys for each new authority and trustees we create for each election. 

Appart from being much more simpler, this make the tests to run faster. In the CI, it tooks 1 minute to run all the app tests, instead of the 14 minutes needed before. 🎉 

This PR also adds new consistency validations between the data received and the signed data.


agustibr (approved):
Great improvement! 🏎😄

slickepinne (approved):
Great improvement! 🤩

beagleknight (approved):
This is huge 🤩 ! Really good job @leio10

PR #69, opened 2021-01-19
Tally endpoint

leio10:
This PR adds the endpoint to perform the tally process.

* It exposes all the `vote` messages to allow the voting scheme to process all the received votes.
* It also allows to hide the `signed_data` until the results are published.
* It implements the tally for the dummy voting scheme and adds a very basic idea of the calls used for the electionguard voting scheme.
* Finally, it add the `start_tally` method to the Ruby client.


decidim-bulletin_board-app/app/commands/process_tally_step.rb
```diff
@@ -0,0 +1,63 @@
...
+  end
+
+  private
+
+  def valid?
+    valid_client?(client.authority? && election.authority == client || client.trustee? && election.trustees.member?(client)) &&
+      valid_author?(message_identifier.from_authority? || message_identifier.from_trustee?) &&
+      valid_step?(election.tally?) &&
+      process_message
+  end
```
beagleknight, line 47: I think this validations are becoming more and more complex. Maybe it's the time to start some kind of `Policy` classes or something like that? We can discuss this... someday, not a priority at the moment 👍

decidim-bulletin_board-app/app/graphql/types/message_interface.rb
```diff
@@ -0,0 +1,18 @@
...
+    include Types::BaseInterface
+
+    field :id, ID, null: false
+    field :election, Types::ElectionType, null: false
+    field :client, Types::ClientType, null: false
+    field :message_id, String, null: false
+    field :signed_data, String, null: true do
+      def resolve(parent, _args, context)
+        log_entry = parent.object
+        log_entry.signed_data if log_entry.visible_for_all? || log_entry.election.authority.api_key == context[:api_key]
```
beagleknight, line 14: This is a nice elegant solution. I like it!

decidim-bulletin_board-app/app/models/concerns/message.rb
```diff
@@ -0,0 +1,17 @@
+# frozen_string_literal: true
+
+require "active_support/concern"
+
+module Message
```
beagleknight, line 5: Not a fan of this name for the concern but I also don't like the `ActsAsStuff` thing. What about something like `IdentifiedByMessageId` or `HasMessageIdentifier` ?

beagleknight (approved):
Great job @leio10 ! I am just missing the changes related to the ruby client in the `CHANGELOG.md` 👍

decidim-bulletin_board-app/app/models/concerns/message.rb
leio10: Yes, I don't like it neither. I used this name because the following reasons, but I'd definitely like to find a longer and more specific name:
* This concern is meant to be used in all the models that represent a message (LogEntry and PendingMessage). 
* We already have a `HasMessageIdentifier` concern in the `commmands` folder (and it's slightly different from this one, so we can't reuse it here).

decidim-bulletin_board-app/app/commands/process_tally_step.rb
leio10: I can think a way to improve its readability very easily. After we merge the open PRs I can add it on a new PR

decidim-bulletin_board-app/app/models/concerns/message.rb
beagleknight: I only can think about `ActsAsMessage`. I blame the ruby community for this! XD

decidim-bulletin_board-ruby/spec/decidim/bulletin_board/authority/start_tally_spec.rb
```diff
@@ -0,0 +1,48 @@
...
+            }
+          }
+        end
+
+        context "when everything is ok" do
+          it "broadcasts ok with the result of the graphql mutation" do
+            expect { subject.call }.to broadcast(:ok)
+          end
+
+          it "uses the graphql client to start the tally and return its result" do
```
slickepinne, line 30:
```suggestion
          it "uses the graphql client to start the tally and returns its result" do
```

slickepinne (approved):
Huge and great PR! 😍 Again, my notes aren't code-related. Amazing work!!

PR #73, opened 2021-01-21
Show some tally messages

leio10:
After merging #69 we detected that the signed data of some tally messages should be accesible for the trustees (e.g. `tally.cast` in the dummy and electionguard schemes).
This PR allow voting schemes to decide which messages should be hidden before the results publishing.
It also splits the `start_tally` from the rest of the tally messages, to improve the consistency of the messages.


decidim-bulletin_board-app/app/services/voting_scheme.rb
```diff
@@ -13,4 +13,8 @@ module VotingScheme
   def self.from_name(name)
     VOTING_SCHEME[name.to_sym]
   end
+
+  def self.results_message?(voting_scheme_name, type_subtype)
+    from_name(voting_scheme_name)::RESULTS.include?(type_subtype)
```
beagleknight, line 18: 👌

beagleknight (approved):
Wow, this is great @leio10 ! Even if we ended with a bit of duplication I think it is totally worth it and the API looks much clearer.

PR #78, opened 2021-01-27
feat: add the IdentificationKeys class to the library

beagleknight:
This PR moves the `IdentificationKeys` code from the application to the JS library. The code itself doesn't need to be reviewed because it was ported from the one that we have on `decidim` with a few small changes. The idea is to be able to remove that same code from `decidim` in favor to keep the implementation details in the library.


decidim-bulletin_board-js/src/trustee/identification_keys.js
```diff
@@ -0,0 +1,256 @@
+/**
+ * IdentificationKeys component.
+ */
+export class IdentificationKeys {
+  constructor(trusteeUniqueId, storedPublicKey) {
+    this.format = "jwk";
+    this.algorithm = {
+      name: "RSASSA-PKCS1-v1_5",
+      modulusLength: 4096,
```
leio10, line 9: It would be great to share this constant with the Ruby code. I think is very hard to do that, but then we should define a constant for that.

leio10 (approved):
Nice! 🔝 🎉

decidim-bulletin_board-js/src/trustee/identification_keys.js
beagleknight: I am going to move it to a constant, and we can think about sharing this with Ruby. Maybe with some kind of JSON configuration file or so.

slickepinne (approved):
👍 Nice!

PR #80, opened 2021-01-28
chore: add timeout to encrypt method

slickepinne:
This PR adds a Timeout of 500ms to the `encrypt` method of the Voter Dummy to be more realistic.


leio10 (approved):
:+1:

orlera (approved):
⏳

PR #83, opened 2021-01-28
feat: Complete the sandbox to have it working with the dummy voting scheme

leio10:
This PR fixes the JS client to make the sandbox to be able to complete a whole election using the Dummy voting scheme.


decidim-bulletin_board-js/src/trustee/trustee.js
```diff
@@ -1,9 +1,9 @@
 /* eslint-disable camelcase */
 import { TrusteeWrapper } from "./trustee_wrapper_dummy";
 import { EventManager } from "./event_manager";
-import { JWTParser } from "./jwt_parser";
+import { JWTParser } from "../jwt_parser";
 
-export const WAIT_TIME_MS = 1_000; // 1s
+export const WAIT_TIME_MS = 100; // 1s
```
beagleknight, line 6: Remove the comment to avoid confusion 👍

decidim-bulletin_board-js/src/voter/voter.js
```diff
@@ -14,22 +15,43 @@ export class Voter {
...
+
   /**
    * Encrypts the data using the wrapper.
    *
    * @param {Object} data - The data to be.
    * @returns {Object} - The data encrypted.
    */
   async encrypt(data) {
-    return this.wrapper.encrypt(data);
+    return await this.wrapper.encrypt(data);
```
beagleknight, line 54: The `await` is not necessary (because the `async` is not necessary neither). You can just return the method call. Also, update the comment because it seems that it is returning a `Promise<Object>`.

decidim-bulletin_board-js/src/voter/voter_wrapper_dummy.js
```diff
@@ -8,11 +13,64 @@ export const WAIT_TIME_MS = 500; // 0.5s
...
+        if (!this.jointElectionKey) {
+          console.warn("Invalid election status.");
+          return;
+        }
+        return JSON.stringify(this.createBallot(vote));
+      }
+    );
+  }
+
+  createBallot(vote) {
```
beagleknight, line 52: I know it is just a dummy wrapper but can you add the documentation blocks that are missing for these methods. Also mark them as private if they are 😄

decidim-bulletin_board-js/src/voter/voter_wrapper_dummy.js
```diff
@@ -8,11 +13,64 @@ export const WAIT_TIME_MS = 500; // 0.5s
...
+          ? 1
+          : 0;
+        contest.ballot_selections.push({
+          object_id: ballotSelection.object_id,
+          ciphertext:
+            voted + Math.floor(Math.random() * 500) * this.jointElectionKey,
+        });
+      }
+      ballot.contests.push(contest);
+    }
```
beagleknight, line 72: I'm not a super fan of using `for` and `push` together. Do you think you could rewrite this using two `map` functions? Something like:

```js
ballot.contests = this.contests.map(contestDescription => {
  return {
    // stuff...
    ballot_selections: contestDespcription.ballot_selections.map(ballotSelection => {
      return {
        // more stuff
      };
    }); 
  }
});
```

I tend to find this functional style a bit more readable. What do you think?

decidim-bulletin_board-ruby/.gitignore
```diff
@@ -15,4 +15,7 @@
 .rubocop-http*
 
 # byebug history
-.byebug_history
\ No newline at end of file
+.byebug_history
+
+# compiled dev JS client
+/app/assets/javascripts/decidim/bulletin_board/decidim-bulletin_board.dev.js
```
beagleknight, line 21: Maybe we can ignore both artifacts since they are just included on the release. We need to ensure that the `gem build` task doesn't ignore them though.

beagleknight (approved):
Great job @leio10 ! This sandbox is getting really shiny 😄 . I just added a couple of suggestions and most of them can be safely ignored and/or implemented in a next PR as needed.

decidim-bulletin_board-ruby/.gitignore
leio10: I've had some problems after deleting the .js, so that's why I've only added the .dev.js, but I agree with you. We can address that in future PRs.

decidim-bulletin_board-js/src/trustee/trustee.js
```diff
@@ -72,7 +72,11 @@ export class Trustee {
    * @returns {Promise<Boolean>}
    */
   async *setupKeyCeremony() {
-    const message = await this.waitForNextLogEntryResult();
+    let message;
+    while (!message) {
+      message = await this.waitForNextLogEntryResult();
+    }
```
beagleknight, line 78: As mentioned before, I am a bit worried about this but since we are dealing with the dummy schema I think is ok for now. We can revisit this code when we add the ElectionGuard wrapper 👍

decidim-bulletin_board-js/src/voter/voter.js
```diff
@@ -14,21 +15,42 @@ export class Voter {
...
+   * @returns {Promise<undefined>}
+   */
+  setup() {
+    return this.bulletinBoardClient
+      .getElectionLogEntries({
+        electionUniqueId: this.electionContext.id,
+      })
+      .then(async (logEntries) => {
+        for (const logEntry of logEntries) {
+          const message = await this.parser.parse(logEntry.signedData);
```
beagleknight, line 41: Do we need to ensure that the `logEntry.signedData` is present?

decidim-bulletin_board-js/src/voter/voter_wrapper_dummy.js
```diff
@@ -8,11 +13,87 @@ export const WAIT_TIME_MS = 500; // 0.5s
...
+    this.contests = {};
+  }
+
+  /**
+   * Process the message and update the wrapper status.
+   *
+   * @param {String} messageId - The identifier of the message.
+   * @param {Object} message - An object with the message to process.
+   *
+   * @returns {Object|undefined}
```
beagleknight, line 26: Just a minor thing: this method always return `undefined`.

decidim-bulletin_board-js/src/voter/voter_wrapper_dummy.js
```diff
@@ -8,11 +13,87 @@ export const WAIT_TIME_MS = 500; // 0.5s
...
+   *
+   * @param {Object} vote - An object with the choosen answers for each question.
+   *
+   * @private
+   * @returns {<Object>}
+   */
+  createBallot(vote) {
+    return {
+      ballot_style: "ballot-style",
+      contests: this.contests.map((contestDescription) => {
```
beagleknight, line 76: 🤩

beagleknight: If you want to save some keystrokes you can apply destructuring:
```js
contests: this.contests.map(({ object_id, ballot_selections }) => {
// ...
```

beagleknight (approved):

decidim-bulletin_board-js/src/voter/voter.js
leio10: We could do it, but now the messages without `signedData` are the tally ones, so there shouldn't be a problem at the moment.

orlera (approved):
🏖️ 🥡

PR #84, opened 2021-01-28
test: install cypress_on_rails and add a simple test as an example

beagleknight:
This setup the e2e test suite for our bulletin board application. It will test our sandbox environment, so we can be sure that both the JS and the Ruby clients work.

I added a simple test as an example, so we can start building something on top of that.

### DIFF
--- a/.github/workflows/app-workflow.yml
+++ b/.github/workflows/app-workflow.yml
@@ -123,3 +123,71 @@ jobs:
         uses: codecov/codecov-action@v1
         with:
           directory: ./decidim-bulletin_board-app/coverage
+  e2e-test:
+    name: E2E tests
+    needs: rspec-test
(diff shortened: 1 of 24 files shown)


.github/workflows/app-workflow.yml
```diff
@@ -123,3 +123,71 @@ jobs:
         uses: codecov/codecov-action@v1
         with:
           directory: ./decidim-bulletin_board-app/coverage
+  e2e-test:
```
leio10, line 126: 😍 🔝 ❤️ 🤩 🥇 💎

leio10 (approved):
This is soooooo good!! 👏👏👏

orlera (approved):
🧷

agustibr (approved):
👏👏

PR #86, opened 2021-01-29
fix(app): remove deprecated babel/polyfill

agustibr:
When deploying the BB app to heroku the `rake assets:precompile` failed because of the `babel/polyfill` package is deprecated this PR removes the dependency and as the babel docs suggest replaces the polyfill with `core-js/stable` and `regenerator-runtime/runtime` -> https://babeljs.io/docs/en/babel-polyfill.

### DIFF
--- a/decidim-bulletin_board-app/app/javascript/packs/application.js
+++ b/decidim-bulletin_board-app/app/javascript/packs/application.js
@@ -3,7 +3,8 @@
 // a relevant structure within app/javascript and only use these pack files to reference
 // that code so it'll be compiled.
 
-import "@babel/polyfill";
+import "core-js/stable";
+import "regenerator-runtime/runtime";
 
 require("@rails/ujs").start();
 require("turbolinks").start();
--- a/decidim-bulletin_board-app/package-lock.json
+++ b/decidim-bulletin_board-app/package-lock.json
@@ -852,24 +852,6 @@
         "@babel/helper-plugin-utils": "^7.10.4"
       }
     },
-    "@babel/polyfill": {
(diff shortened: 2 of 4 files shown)


beagleknight (approved):
Great! 🤩 . At some point we should get rid off one lock file. Maintaining two is not that great 😅

agustibr:
@beagleknight I've attempted to remove it, but as far as I understand we should switch GithubActions to use `yarn` instead of `npm`.

Because when deleting the `package-lock.json` and running the `npm ci` commands raises this error:
```
% npm ci
npm WARN prepare removing existing node_modules/ before installation
npm ERR! cipm can only install packages with an existing package-lock.json or npm-shrinkwrap.json 
with lockfileVersion >= 1. Run an install with npm@5 or later to generate it, then try again.
```

beagleknight:
> @beagleknight I've attempted to remove it, but as far as I understand we should switch GithubActions to use `yarn` instead of `npm`.
> 
> Because when deleting the `package-lock.json` and running the `npm ci` commands raises this error:
> 
> ```
> % npm ci
> npm WARN prepare removing existing node_modules/ before installation
> npm ERR! cipm can only install packages with an existing package-lock.json or npm-shrinkwrap.json 
> with lockfileVersion >= 1. Run an install with npm@5 or later to generate it, then try again.
> ```

It's the other way around! We need to use `npm` everywhere. By default, Rails come with `yarn` enabled, but I think that it's not needed at all. Don't worry about it for now we can deal with it in a future PR, and we can discuss it with the team, of course!

agustibr:
😅

PR #90, opened 2021-02-02
Fix BulletinBoard version and graphiql path in `api/docs`

agustibr:
This PR fixes the BB version showed in the `api/docs` page, it also fixes the path to `graphiql`.

Screenshot 2021-02-02 at 10 52 30


leio10 (approved):
Good catch! ❤️

orlera (approved):

PR #94, opened 2021-02-04
test: add a single e2e test for the whole election process

beagleknight:
This continues with E2E testing. We discussed with the team that it is better to have a single test that perform all the steps from an election created to the end. I've changed the existing tests and refactored them to a single one.

At the moment we are testing everything until the ending the vote. You can see it in action below:

video


decidim-bulletin_board-app/app/assets/javascripts/sandbox/vote.es6
```diff
@@ -9,6 +9,7 @@ $(async () => {
   const $castVote = $voter.find("button");
   const $vote = $voter.find("textarea");
   const $voterId = $voter.find("input");
+  const $doneMessage = $voter.find(".done-message");
 
   $vote.on("change", (event) => {
     $vote.css("border", "");
```
leio10, line 15: as we now have a css file, maybe it would be better to use classes for these things

decidim-bulletin_board-app/app/assets/javascripts/sandbox/vote.es6
```diff
@@ -29,13 +30,16 @@ $(async () => {
...
-    onSetup() {},
+    onSetup() {
+      $castVote.removeAttr("disabled");
+    },
     onBindStartButton(onEventTriggered) {
       $castVote.on("click", onEventTriggered);
     },
     onStart() {},
     onVoteValidation(validVoteFn, invalidVoteFn) {
       $vote.css("background", "");
```
leio10, line 41: same as before

decidim-bulletin_board-app/app/assets/javascripts/sandbox/vote.es6
```diff
@@ -60,6 +64,7 @@ $(async () => {
       });
     },
     onComplete() {
+      $doneMessage.show();
       $vote.css("background", "green");
```
leio10, line 68: same as before

leio10 (approved):
Well done! 🚜 👍

decidim-bulletin_board-app/app/assets/javascripts/sandbox/vote.es6
beagleknight: It's not the same as showing or hiding something. In this case, this is disabling the button, so it's not just aesthetics 😄

decidim-bulletin_board-app/app/assets/javascripts/sandbox/vote.es6
beagleknight: > as we now have a css file, maybe it would be better to use classes for these things

I agree with your suggestion, but I prefer doing refactors in a separate test just to keep the PR for a single purpose. Actually, having a good e2e test suite will make those refactor a breeze 😄

decidim-bulletin_board-app/spec/cypress/integration/sandbox/election.page.js
```diff
@@ -0,0 +1,222 @@
...
+   */
+  assertVoteHasEnded(election) {
+    this.withinElectionRow(election, () => {
+      cy.findByText("vote").should("not.exist");
+      cy.findByText("vote_ended").should("be.visible");
+    });
+  }
+
+  /**
+   * Prepares a context to find elements in a election row from the list.
```
slickepinne, line 212:
```suggestion
   * Prepare a context to find elements in an election row from the list.
```
Here I just removed the s so it's consistent with the rest of the comments.

slickepinne (approved):
Wonderful!

decidim-bulletin_board-app/spec/cypress/integration/sandbox/election.page.js
beagleknight: > Here I just removed the s so it's consistent with the rest of the comments.

I'm going to fix it in #95 👍 , thanks!

PR #95, opened 2021-02-04
test: add the tally process to the e2e test

beagleknight:
This is built on top of #94, and it adds the tally process to the whole test. I also refactored some things to make the test less flaky 👍 

video

### DIFF
--- a/.github/workflows/app-workflow.yml
+++ b/.github/workflows/app-workflow.yml
@@ -192,16 +192,19 @@ jobs:
       - name: Run tests
         run: npm run e2e:tests
       - name: Upload artifacts (screenshots)
+        if: ${{ always() }}
         uses: actions/upload-artifact@v2
         with:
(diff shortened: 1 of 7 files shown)


agustibr (approved):

.github/workflows/app-workflow.yml
```diff
@@ -192,16 +192,19 @@ jobs:
       - name: Run tests
         run: npm run e2e:tests
       - name: Upload artifacts (screenshots)
+        if: ${{ always() }}
```
leio10, line 195: what does this does? 😅

leio10 (approved):
👏 👏

.github/workflows/app-workflow.yml
beagleknight: This uploads the artifacts even if the test fails 👍

PR #96, opened 2021-02-04
Fix dummy schema results calculations & other fixes

leio10:
This PR implements some fixes to the BB app:
* It fixes the calculation of the results of an election with the dummy voting scheme.
* It allows to enqueue several pending messages for the same client, if they don't have the same `message_id`.
* It fixes the seeding process to use the right election keys for the trustees and to sign correctly the messages sent by the BB.
* It adds the `decodedData` field to the graphQL interface to ease the debugging of the messages.

### DIFF
--- a/decidim-bulletin_board-app/app/services/voting_scheme/dummy.rb
+++ b/decidim-bulletin_board-app/app/services/voting_scheme/dummy.rb
@@ -122,7 +122,7 @@ module VotingScheme
       results = build_questions_struct(0)
       state[:joint_shares].each do |question, answers|
         answers.each do |answer, joint_share|
-          results[question][answer] = (joint_share / state[:joint_election_key])**(1.0 / state[:trustees].count).round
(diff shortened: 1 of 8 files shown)


decidim-bulletin_board-app/app/services/voting_scheme/dummy.rb
```diff
@@ -122,7 +122,7 @@ def process_tally_message(message_identifier, _message, content)
       results = build_questions_struct(0)
       state[:joint_shares].each do |question, answers|
         answers.each do |answer, joint_share|
-          results[question][answer] = (joint_share / state[:joint_election_key])**(1.0 / state[:trustees].count).round
+          results[question][answer] = ((joint_share / state[:joint_election_key])**(1.0 / state[:trustees].count)).round
```
orlera, line 125: This looks like one of those 🐛 s that can take hours to identify 😄 
Nice catch! 🗜️

orlera (approved):
💎

agustibr (approved):

PR #97, opened 2021-02-04
Send election results

slickepinne:
This PR renames the `get_election_log_entries_by_types` command to `get_election_results`. With the changes made in this command, the decidim instance doesn't need to decode the `signed_data` of the `log_entry` but receives the decoded results directly.


decidim-bulletin_board-ruby/lib/decidim/bulletin_board/authority/get_election_results.rb
```diff
@@ -3,15 +3,13 @@
 module Decidim
   module BulletinBoard
     module Authority
-      # This command uses the GraphQL client to get the log entries of an election.
-      class GetElectionLogEntriesByTypes < Decidim::BulletinBoard::Command
+      # This command uses the GraphQL client to get the the results of an election.
```
orlera, line 6:
```suggestion
      # This command uses the GraphQL client to get the results of an election.
```

orlera: Sweet revenge 🎵

decidim-bulletin_board-ruby/lib/decidim/bulletin_board/authority/get_election_results.rb
```diff
@@ -37,14 +37,24 @@ def call
             end
           end
 
-          broadcast(:ok, response.data.election.log_entries)
+          @signed_data = response.data.election.log_entries.first.signed_data
```
orlera, line 40: Genuine question out of ignorance:

In case we called this command when the election does NOT yet have the `end_tally` log entry, would the above GraphQL query raise an error that will be rescued or will we get an empty `response.data.election.log_entries` array and fail with an uncaught exception here?

orlera (approved):
Nice job! 🐦 
Added a question that _might_ result in a change 😄

decidim-bulletin_board-ruby/lib/decidim/bulletin_board/authority/get_election_results.rb
slickepinne: Thanks a lot for that!!

leio10 (approved):
Amazing work!! 🔝 👏

PR #102, opened 2021-02-11
Verify JWT tokens on the JS clients (trustees and voters)

leio10:
This PR refactors the `JWTParser` class to implement a verification of the received tokens with the `MessageParser` class. It uses the `node-jose` library to do that. This class also parse the `messageId` string into a `MessageIdentifier` structure.

To have the sandbox working with these changes, it also updates the seeds with the last `create_election` message changes and it uses the `types` filter for the election log entries retrieval (to avoid processing vote messages during the tally).


decidim-bulletin_board-js/src/client/message-parser.js
```diff
@@ -0,0 +1,99 @@
...
+   *
+   * @constructor
+   * @param {Object} params - An object that contains the initialization params.
+   * - {String} authorityPublicKeyJSON - The public key of the authority
+   */
+  constructor({ authorityPublicKeyJSON }) {
+    this.authorityPublicKeyJSON = authorityPublicKeyJSON;
+    JWK.asKey(authorityPublicKeyJSON, "json").then((result) => {
+      this.authorityPublicKey = result;
+    });
```
beagleknight, line 25: This is a bit "dangerous" since you are assigning that value when the promise is resolved and you don't know when that will be. I propose you an alternative 👍 

```js
this.authorityPublicKey = JWK.asKey(authorityPublicKeyJSON, "json");
```

Then, when you want to use that in the code you do something like this:

```js
async myFunction() {
  // ...
  const authorityPublicKey = await this.authorityPublicKey;
  // ...
}
```

This will make sure that you have an actual value to work with 😄

decidim-bulletin_board-js/src/client/message-parser.js
```diff
@@ -0,0 +1,99 @@
...
+
+  /**
+   * Parses the given token only if it can be verified.
+   *
+   * @param {String} token - A JWT token.
+   * @returns {Promise<Object>} - The payload included in the token.
+   * @throws An error is thrown if the payload is not a valid JSON or the token
+   *         cannot be verified.
+   */
+  async parse({ messageId, signedData }) {
```
beagleknight, line 37: If you don't need the `messageId` I recommend just passing the `messageIdentifier` to the function to remove the dependency. What do you think?

decidim-bulletin_board-js/src/client/message-parser.js
```diff
@@ -0,0 +1,99 @@
...
+    );
+    for (const trustee of trustees) {
+      promises.push(
+        this.loadKey(trustee).then((key) => {
+          result[TRUSTEE_TYPE][trustee.slug] = key;
+        })
+      );
+    }
+
+    await Promise.all(promises);
```
beagleknight, line 90: Just a personal preference but consider writing this using `async/await` to make it a bit clearer.

```js
const bulletinBoardKey = await this.loadKey(bulletin_board);
result[BULLETIN_BOARD_TYPE][bulletin_board.slug] = key;
// and so on...
```

beagleknight (approved):
Great job ✨ ! Feel free to implement any feedback in a future PR, nothing blocking here!

decidim-bulletin_board-js/src/client/message-parser.js
leio10: I did it that way at the beginning, but I had problems when doing it inside the `for` loop, because the end was reached before completing all the loads 😕. 

I'm not completely sure, but I think that this way it's "starting" all the processes and then wait for all the promises to end. I'm afraid that if we use the `await` we would have to wait for each key for being parsed to start the next one.

decidim-bulletin_board-js/src/client/message-parser.js
leio10: mm, I think that this is the best place to do it, as you always need to parse the `messageId` when you are parsing the `signedData`. If we move this to the caller, we would have to copy that line to the trustee and to the voter.

decidim-bulletin_board-js/src/client/message-parser.js
leio10: Thanks! I had this pending to be resolved and your solution is exactly what I needed 😍

decidim-bulletin_board-js/src/client/message-parser.js
beagleknight: Well even you use `Promise.all` internally all promises are being resolved one after another. There is no parallelism in JS 😄 . As I said your solution is correct but in terms of performances is the same as just using `await` in a for loop 👍

PR #103, opened 2021-02-11
Allow to create elections on the sandbox

leio10:
This PR adds the `create_election` command to be tested in the sandbox. It also changes the election title field to a `jsonb` column to be able to store different translations for that text.


slickepinne (approved):
Nice!!! 😍

beagleknight (approved):
:sparkles:

PR #104, opened 2021-02-12
refactor: move js library to the application

beagleknight:
After discussing it with the team we realized that the development workflow was not so great, so we decided to make some changes.

First, the `decidim-bulletin_board-js` folder has been removed from the project and all the source files have been moved to the bulletin board application. You can now find them in `decidim-bulletin_board-app/app/javascript/decidim-bulletin_board`. The application has been updated to use Webpack 6, and it is going to use the assets in development when we are working in the sandbox. If you want to work in the sandbox you need to open two terminals now and run these commands:

```
$ bundle exec rails s    # Start rails server as usual
$ bin/webpack --watch    # Start the webpack build in watch mode
```

You may notice that the bulletin board application is now using the gem from the local file system now. That means the sandbox environment will be targeting the development branch always.

Finally, we added a `release` script in the `decidim-bulletin_board-ruby` folder. After you updated the version and the `CHANGELOG` accordingly you can just run `bin/release` to release a new version of the gem. That will compile the assets from the application and release the gem as usual.


.devcontainer/Dockerfile
```diff
@@ -19,8 +19,5 @@ RUN apt-get install -y nodejs
 # Install Rails
 RUN gem install bundler
 
-# Install Yarn
```
slickepinne, line 22: 🗑 😂

slickepinne (approved):
Really, really nice. And goodbye `yarn`. 👋 🚮

leio10 (approved):
Gooood!! 🚀

PR #105, opened 2021-02-15
Use consistent and more explicit names for settings

leio10:
This PR changes the settings for the ruby BB client and the app to improve code readability. It uses prefixes for all BB and authority settings (names, public and private keys, etc).

It also removes the need to use a setting to define the API endpoint for the sandbox.

Finally, it improves the vote sandbox page UX a bit.


decidim-bulletin_board-app/app/javascript/packs/sandbox/vote.js
```diff
@@ -97,6 +97,8 @@ $(async () => {
       });
     },
     onComplete() {
+      $castVote.prop("disabled", true);
```
slickepinne, line 100: Oh thanks, I forgot about that! 💙

slickepinne (approved):
So nice! 🌻

beagleknight (approved):
Consistency overflow ✨ . Great job!

PR #106, opened 2021-02-16
Use the trustee wrapper adapter pattern

beagleknight:
This PR removes the "dummy" voting scheme dependency from the project and put the JS code in a separate folder called `voting-scheme-dummy`.

The idea is when an application wants to use any component (i.e. `KeyCeremonyComponent`), they will need to provide a `wrapperAdapter`. That's a simple JS class that will interact with the corresponding wrapper to perform the necessary operations. A `TrusteeWrapperAdapter` abstract class has been created to make this easier. The idea is users can copy that interface and implement their own methods as we did for the "dummy".

After this we will create a `voting-scheme-election_guard` to add the Electionguard implementation 👍


leio10 (approved):
Great refactor! The project is prepared to add the electionguard wrapper now!! 💪 🤩

orlera (approved):
👾

PR #107, opened 2021-02-17
Setup election_guard voting scheme project

beagleknight:
This adds a new folder `voting-scheme-election_guard` that includes the `TrusteeWrapperAdapter` for the Electionguard voting scheme.

At this moment we cannot test that in the e2e, but we can do it manually. You need to create a new election in the sandbox and then perform the key ceremony. It will take a while because we are using the `pyodide` build for the wrapper, so it uses web workers and web assembly to do the heavy computation. In my computer and running three trustees at the same time it took almost an hour. It is a bit random, so it can be faster in your machine.

There are a few things to consider at the moment:
- You should wait until the page has loaded all the python stuff. I am going to add a `setup` method for the wrapper soon but as a workaround just wait a bit 😄 
- The `voting-scheme-election_guard` folder includes the `pyodide` build in the `vendor` folder. These files needs to be statically serve by the bulletin board, so I just added a symbolic link to the bulletin board vendor folder. In decidim we will need to do something similar.
- The `restore` doesn't work at the moment.


orlera (approved):
🎉

agustibr (approved):
😄 wohooo, great!

leio10 (approved):
Awesome!! 🤩 💪

PR #108, opened 2021-02-17
Add election_guard-js project

beagleknight:


beagleknight:
We are still thinking the good approach for this so I am going to close it to keep things tidy.

PR #109, opened 2021-02-18
Add the VoterWrapperAdapter for the dummy scheme

beagleknight:
This applies the same pattern that we saw in #106. The idea is to extract the dummy voting scheme part from the voter and use an adapter. Then we will be able to do the same with `election_guard`.


leio10 (approved):
Willing to see next PR!! 😍

orlera (approved):
🚂

agustibr (approved):

voting-scheme-dummy/src/voter_wrapper_adapter.js
```diff
@@ -0,0 +1,43 @@
...
+   * @param {Object} params - An object that contains the initialization params.
+   * - {String} voterId - The unique id of a trustee.
+   */
+  constructor({ voterId }) {
+    this.voterId = voterId;
+    this.wrapper = new VoterWrapper({ voterId });
+  }
+
+  /**
+   * Process the message and update the wrapper status.
```
slickepinne, line 21:
```suggestion
   * Processes the message and updates the wrapper status.
```

slickepinne (approved):
Amazing!

voting-scheme-dummy/src/voter_wrapper_adapter.js
beagleknight: Discussed on slack to apply them in #110

PR #110, opened 2021-02-18
Get ballot hash before auditing or casting

slickepinne:
With this PR, the `ballot_hash` is available after `encrypting` the vote. Therefore, it can be displayed before deciding to `audit` or `cast` a vote.

### DIFF
--- a/decidim-bulletin_board-app/cypress/integration/sandbox/election.page.js
+++ b/decidim-bulletin_board-app/cypress/integration/sandbox/election.page.js
@@ -141,13 +141,31 @@ export class ElectionPage {
   }
 
   /**
+   * Encrypt a vote.
+   *
+   * @returns {undefined}
+   */
+  encryptVote() {
+    cy.findByText("Vote").click();
+    cy.findByText("Encrypt vote").should("be.visible").click();
+  }
+
+  /**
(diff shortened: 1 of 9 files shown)


agustibr (approved):
👍

beagleknight (approved):
Good job 🤩 !

decidim-bulletin_board-app/cypress/integration/sandbox/election.page.js
```diff
@@ -141,13 +141,31 @@ export class ElectionPage {
...
+  }
+
+  /**
+   * Assert that the ballot hash is present
+   *
+   * @returns {undefined}
+   */
+  assertBallotHashIsPresent() {
+    cy.findByText("Encrypt vote").should("be.disabled");
+    cy.get(".ballot-hash").should("include.text", "Your ballot identifier is:");
```
beagleknight, line 160: Avoid using selectors if possible. I think it's better to just use `cy.findbyText("Your ballot identifier is:").should("be.visible")`.

beagleknight (approved):
Great!

leio10 (approved):
Great worked! I love the new names and the responsibilities reorganization ❤️ 😻

PR #113, opened 2021-02-22
Add the VoterWrapperAdapter for the election_guard scheme

beagleknight:
This applies the same pattern that we saw in #109. The idea is to extract the `election_guard` voting scheme part from the voter and use an adapter.


orlera (approved):

slickepinne (approved):
Great!

PR #114, opened 2021-02-22
Update README with sandbox instructions

slickepinne:
This PR adds the two (_actually, it's only one besides the normal start of the server_) commands to the app README.


agustibr (approved):
Thanks

orlera (approved):

leio10 (approved):
🥇

PR #115, opened 2021-02-22
Add election_guard integration for the tally process

beagleknight:
This changes the tally process to support the `election_guard` wrappers as well. With this PR we can celebrate the whole election process using the two voting schemes.


orlera (approved):
🚀 ❤️ 🌔 🧑‍🚀

leio10 (approved):
Awesome!! 🤠 🤟

PR #116, opened 2021-02-23
Activate sandbox if required

leio10:
Allow to set an environment variable to use the sandbox on the staging server.


slickepinne (approved):
Coool! 👍

PR #118, opened 2021-02-23
Convert dicts to hashes recursively

leio10:
This PR fixes a problem we had when processing the electionguard wrappers results. When we get nested `dict`s from python, it was only converting to hash the main object, but not the nested `dict`s. With this changes, this conversion is only done in the `electionguard` service, and it's done recursively.


slickepinne (approved):
🙇🏻‍♀️ 😄

beagleknight (approved):

PR #119, opened 2021-02-23
fix: wait until cast succeed to call the correct callback

beagleknight:
We need to wait until the `onCastBallot` has finished before showing the cast complete message.


leio10 (approved):
:+1:

slickepinne (approved):
Lovely!

agustibr (approved):

PR #120, opened 2021-02-24
fix: start tally must return a hash

beagleknight:
All the processed messages must return a hash and we made a mistake with the `start_tally` one.


slickepinne (approved):

leio10 (approved):

orlera (approved):
#️⃣

PR #121, opened 2021-02-24
Add a Dockerfile to deploy the application in heroku using the container stack

beagleknight:
I added a `Dockerfile` so we can deploy the application everywhere 👍


orlera (approved):
🐳

leio10 (approved):

Dockerfile
```diff
@@ -0,0 +1,60 @@
+FROM ruby:2.6.6
```
oriolgual, line 1: Why not 2.7?

beagleknight: Because we are using `2.6.6` in the Gemfile at the moment and the `Dockerfile` should use the same 👍

PR #122, opened 2021-02-24
Fix sequence order and hiding decodedData for results

leio10:
This PR fixes two issues:
* For ElectionGuard, the `sequenceOrder` cannot have the same value for two questions or two answers in the same question. Instead of using the weight, this PR makes the create election to use the `sequenceOrder` based on the order in the list where the questions and answers are in the `electionData` parameters.
* In the GraphQL API, when the `signedData` was hidden, the `decodedData` were still visible. This PR fixes that.


beagleknight (approved):
✨

orlera (approved):

PR #124, opened 2021-02-25
Replace message_type with message_id on the BB messages

leio10:
This PR adds the `message_id` and removes the `message_type` from the `signedData` field on the BB messages. This enhances the consistency between different messages.


beagleknight (approved):
👍

orlera (approved):

PR #129, opened 2021-02-26
Wrappers return a list of messages instead of a single message

txus:
* Depends on https://github.com/decidim/decidim-electionguard/pull/11

### DIFF
--- a/decidim-bulletin_board-app/app/commands/process_key_ceremony_step.rb
+++ b/decidim-bulletin_board-app/app/commands/process_key_ceremony_step.rb
@@ -13,7 +13,7 @@ class ProcessKeyCeremonyStep < Rectify::Command
     @client = @trustee = trustee
     @message_id = message_id
     @signed_data = signed_data
-    @response_log_entry = nil
+    @response_log_entries = nil
   end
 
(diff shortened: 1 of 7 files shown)


decidim-bulletin_board-app/app/commands/process_key_ceremony_step.rb
```diff
@@ -44,10 +44,10 @@ def call
 
   private
 
-  attr_accessor :trustee, :response_log_entry
+  attr_accessor :trustee, :response_log_entries
 
   def save_election!
-    if response_log_entry && response_log_entry.message_identifier.type == "end_key_ceremony"
+    if response_log_entries && response_log_entries.any? { |le| le.message_identifier.type == "end_key_ceremony" }
```
leio10, line 50: I guess that here and in the tally we can assume that the `end_...` will be the last response, right? Even with that, maybe this approach is healthier. 🤔

leio10 (changes requested):
Nice!! 😍 👏

beagleknight (approved):

leio10 (approved):
👍

PR #134, opened 2021-03-03
Add basic file processing for the election

beagleknight:
This adds the first check on the election log verifier. You can test it like this:

```bash
bin/verify test/fixtures/election-ok.tar
bin/verify test/fixtures/test/fixtures/election-wrong-chained-hashes.tar
```

In this PR I also keep building the skeleton of the verifier app, so it can change in the future to improve performance and so.


verifier/src/ballot/index.js
```diff
@@ -1,8 +1,18 @@
...
+  /**
+   * Parse the audit ballot and run some verifiers.
+   *
+   * @param {String} path - an audit ballot file path.
+   * @returns {Promise<undefined>}
+   */
   verify() {
-    console.log(`Ballot is ${chalk.green("OK")}.`);
-    process.exit(0);
+    console.log(`${chalk.yellow("Verifying ballot...")}`);
```
orlera, line 16: 🎨

verifier/src/election/verifiers/chained-hash.verifier.js
```diff
@@ -0,0 +1,53 @@
...
+    return new Promise((resolve) => {
+      readAllElectionFiles(
+        path,
+        (data, fail) => {
+          const { signed_data: signedData, chained_hash: chainedHash } = data;
+          // The first hash is computed using the election unique id.
+          if (!previousHash) {
+            const { message_id: messageId } = data;
+            const [authorityName, electionId] = messageId.split(".");
+            previousHash = [authorityName, electionId].join(".");
```
orlera, line 31: I must be not seeing something somewhere, but this to me looks like
```
previousHash = messageId
```

orlera (approved):
Nice 🤩

verifier/src/election/verifiers/chained-hash.verifier.js
beagleknight: @orlera The `messageId` is longer 😅  so I am just taking the first two elements!

verifier/src/jwt-utils.js
```diff
@@ -0,0 +1,28 @@
+/**
```
leio10, line 1: Why don't we use a jwt library, as in the app?

leio10 (approved):
Nice work!! 🧶 🐈

agustibr (approved):

verifier/src/jwt-utils.js
beagleknight: Oh! I thought that at the moment I only need to decode it, and it is fairly simple without external libraries. When I need to check signatures I'll add it for sure 😄 !

PR #135, opened 2021-03-03
Add verification results file generation

leio10:
This PR adds the generation of the verifiable results file after publishing the results. To do that, it converts the `publish_results` mutation to an asynchronous task (as the rest of the mutation, except `create_election`). 

This verifiable results file is a `tar` file, with one `.jsonl.gz` (GZip compressed file with multiple JSON lines) file for each possible status for the election, including all the log entries received when the election was on that status. The messages that make the election to change the status are included in the target status file. This is the list of files included in a typical verifiable results file:
* 0-created.jsonl.gz
* 1-key_ceremony.jsonl.gz
* 2-key_ceremony_ended.jsonl.gz
* 3-vote.jsonl.gz
* 4-vote_ended.jsonl.gz
* 5-tally.jsonl.gz
* 6-tally_ended.jsonl.gz
* 7-results_published.jsonl.gz

Each line of the files consists on a JSON document representing an election log entry with four keys: `message_id`, `signed_data`,  `chained_hash` and `content_hash`.

This PR also fixes the datatype of the log entries's `content_hash`, that was `text`, where `string` is enough.


orlera (approved):
Nice! Neat solution 🧼

beagleknight (approved):
This is great ✨ . Coding-wise nothing more to add but as a future improvement we could move the file generation to another command or service to keep the `publish_results` command easier to test and maintain.

PR #136, opened 2021-03-03
Generate bulk votes from sandbox - dummy schema

orlera:
Allows generating a user-defined number of votes encrypted with the dummy schema, replicating https://github.com/decidim/decidim-bulletin-board/blob/develop/voting_schemes/dummy/js-adapter/src/voter_wrapper.js#L78.

The votes are stored in a CSV file that will later be used by JMeter.

For the moment being, the logic is in the controller and does not check what schema the election uses.
When introducing support to the EG schema these things will be addressed.

In my local machine, it takes about 20 second per 1000 votes.

[bulk votes](https://user-images.githubusercontent.com/5033945/109857627-2fc15680-7c5b-11eb-8ed1-53e3cbf6c82d.mov)

### DIFF
--- a/decidim-bulletin_board-app/app/controllers/sandbox/elections_controller.rb
+++ b/decidim-bulletin_board-app/app/controllers/sandbox/elections_controller.rb
@@ -33,6 +33,14 @@ module Sandbox
       bulletin_board_client.cast_vote(election_id, params[:voter_id], params[:encrypted_ballot])
     end
 
+    def generate_bulk_votes
+      number_of_votes_to_generate.times do
+        file_client.cast_vote(election_id, SecureRandom.hex, random_encrypted_vote)
+      end
+
+      go_back
+    end
+
     def end_vote
       bulletin_board_client.end_vote(election_id)
       go_back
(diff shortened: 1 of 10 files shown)


agustibr (approved):
🚚 🚚

slickepinne (approved):
Looks amazing!

decidim-bulletin_board-app/app/javascript/packs/sandbox/index.js
```diff
@@ -0,0 +1,37 @@
...
+      contentType: "application/json",
+      data: JSON.stringify({
+        number_of_votes: votesToGenerate,
+      }), // eslint-disable-line camelcase
+      headers: {
+        "X-CSRF-Token": $("meta[name=csrf-token]").attr("content"),
+      },
+    })
+      .done(() => $(event.target).html(`${votesToGenerate} votes generated!`))
+      .fail(() => $(event.target).html(`Failed, retry!`));
```
beagleknight, line 35: Since `$.ajax` returns a `Promise` you could also write this using `async/await` like this:

```js
$generateVotesButton.on("click", async (event) => {
  // ...
  try {
    await $.ajax(...);
    $(event.target).html(`${votesToGenerate} votes generated!`);
  } catch {
    $(event.target).html(`Failed, retry!`);
  }
});
```

I am ok with both ways :)

beagleknight (approved):
Nice ✨ !

decidim-bulletin_board-app/app/controllers/sandbox/elections_controller.rb
```diff
@@ -171,5 +179,37 @@ def base_vote
...
+        end
+      }.to_json
+    end
+
+    def random_answer(max_reached)
+      max_reached ? 0 : rand(0..1)
+    end
+
+    def joint_election_key
+      @joint_election_key ||= JSON.parse(election.log_entries.where(message_type: "end_key_ceremony").last.decoded_data["content"])["joint_election_key"]
```
leio10, line 223: 😻

leio10 (approved):
Great work! 💪 ❤️

PR #139, opened 2021-03-05
Election fixes: chained_hash generation and election creating

leio10:
This PR fixes the generation of the chained hashes. We were using the `decoded_data`, which is a ruby hash of hashes (in this case "hash" as dictionary or hashmap, not as string digest), instead of using the `signed_data`, which is a string and much more easy to be calculated equal on different environments.

It also avoid problems when retrieving the previous chained hash when log entries are cached within the election, that were happening during the seed process.

Finally, during the creation of an election it uses the `max_selections` value for the number of answers to select instead of using the total count of answers.

### DIFF
--- a/decidim-bulletin_board-app/app/models/election.rb
+++ b/decidim-bulletin_board-app/app/models/election.rb
@@ -3,7 +3,7 @@
 class Election < ApplicationRecord
   belongs_to :authority
   has_many :election_trustees
-  has_many :trustees, through: :election_trustees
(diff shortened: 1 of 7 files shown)


orlera (approved):
👍

agustibr (approved):

slickepinne (approved):
Great! 👍

beagleknight (approved):
Nice 👍

PR #140, opened 2021-03-05
Add download button for bulk votes

orlera:
Adds the possibility to download the generated votes CSV from a button in the sandbox.

**Disclaimer**: Heroku timeouts requests over 30s, so it's not possible to generate more than ~1000 votes with the current method.
For the moment I'll just manually merge several files, but if we want to automatise the load test we'll need to move the file generation to a background job.

[](https://user-images.githubusercontent.com/5033945/110104401-67d5b000-7da7-11eb-8f70-2c88b18f1e53.mov)

Also: moves `require "test/private_keys"` to the sandbox controller to have it available in all environments

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -22,6 +22,9 @@
 /decidim-bulletin_board-app/storage/*
 !/decidim-bulletin_board-app/storage/.keep
 
+# Ignore generated votes file
+/decidim-bulletin_board-app/bulk_votes.csv
+
 /decidim-bulletin_board-app/public/assets
 /decidim-bulletin_board-app/.byebug_history
 
--- a/.gitignore
+++ b/.gitignore
@@ -22,9 +22,6 @@
 /decidim-bulletin_board-app/storage/*
 !/decidim-bulletin_board-app/storage/.keep
 
-# Ignore generated votes file
-/decidim-bulletin_board-app/bulk_votes.csv
-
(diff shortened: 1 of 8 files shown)


.gitignore
```diff
@@ -22,6 +22,9 @@
 /decidim-bulletin_board-app/storage/*
 !/decidim-bulletin_board-app/storage/.keep
 
+# Ignore generated votes file
+/decidim-bulletin_board-app/bulk_votes.csv
```
beagleknight, line 26: What about generating that file on the `tmp` folder? Do you think is feasible?

beagleknight (approved):
Great work! ✨ Just check my suggestion in case is something that we can do easily. Otherwise feel free to skip it!
About the heroku timeout I think we can setup that using some environment variable.

orlera:
> About the heroku timeout I think we can setup that using some environment variable.

@beagleknight quoting the [Heroku docs](https://devcenter.heroku.com/articles/request-timeout)
```
The timeout value is not configurable.
If your server requires longer than 30 seconds to complete a given request,
we recommend moving that work to a background task or worker
to periodically ping your server to see if the processing request has been finished.
This pattern frees your web processes up to do more work, and decreases overall application response times.
```

leio10 (approved):
👍

PR #141, opened 2021-03-05
Compile python EG wrappers to JS using pyodide

leio10:
This PR adds the changes from https://github.com/codegram/pyodide to compile the Python wrappers into JS code. 

The command to run the compilation process is `voting_schemes/election_guard/python-to-js/build`. This process needs to have Docker installed. If the script is executed outside the VS dev container, you should create a temporary folder and use the absolute path of that folder as a parameter to the script.

It also replaces `election_guard` to `electionguard` everywhere, and changes the name of the python package from `decidim.electionguard` to `bulletin_board.electionguard`.

Finally, it modifies the EG js code to work with all the last changes made on the python wrappers code.


decidim-bulletin_board-app/cypress/integration/sandbox/election.spec.js
```diff
@@ -56,11 +56,11 @@ describe("Election", () => {
     });
   });
 
-  it("complete the whole process using the election_guard voting scheme", () => {
+  it("complete the whole process using the electionguard voting scheme", () => {
     testElection({
-      electionTitle: "My election_guard election",
-      electionUniqueId: "election_guard-1",
-      electionVotingSchemeName: "election_guard",
+      electionTitle: "My ElectionGuard election",
```
orlera, line 61: It looks like this has been camel-cased on purpose, right?
Just wondering if it was a find&replace issue, that might have had an impact in more critical spots (haven't spotted any other inconsistency, but I'm only human, after all 🤖 ).

orlera (approved):

decidim-bulletin_board-app/cypress/integration/sandbox/election.spec.js
leio10: Mmm, you are right that this is not consistent with the rest of the changes, but it shouldn't affect anything in this case, as this is just a regular string. 

The main problem is that Microsoft uses `ElectionGuard` for the title of the project but `electionguard` for naming all identifiers (projects, namespaces, etc). So, I'm trying to keep that naming schema, but Rails will search for `election_guard` when you define an `ElectionGuard` class. That was why it was done as it was. Then, when mixing Ruby and Python code, it was very hard to keep it consistent. I think that the less worst solution is to keep using  `electionguard` everywhere, and use `Electionguard` for class names on Ruby.

txus (approved):

decidim-bulletin_board-app/app/services/voting_scheme/electionguard.rb
```diff
@@ -2,11 +2,11 @@
 
 require "pycall/dict"
 require "pycall"
-EGBB = PyCall.import_module("decidim.electionguard.bulletin_board")
+EGBB = PyCall.import_module("bulletin_board.electionguard.bulletin_board")
 
 module VotingScheme
   # An implementation of the ElectionGuard voting scheme, using PyCall to execute the ElectionGuard python code
```
slickepinne, line 8:
```suggestion
  # An implementation of the Electionguard voting scheme, using PyCall to execute the Electionguard python code
```

voting_schemes/electionguard/js-adapter/src/wrapper_adapter.js
```diff
@@ -1,5 +1,5 @@
 /**
- * Common class used by any election_guard adapter to run python code through web workers.
+ * Common class used by any ElectionGuard adapter to run python code through web workers.
```
slickepinne, line 2:
```suggestion
 * Common class used by any Electionguard adapter to run python code through web workers.
```

voting_schemes/electionguard/python-to-js/packages/jsons/meta.yaml
```diff
@@ -0,0 +1,11 @@
+package:
+  name: jsons
+  version: 1.2.0
+source:
+  sha256: 7ed8ea50dff03bef3ed44c4ea34bb72015fbdff02d069e2f9e81eb67224e1514
+  url: file:dist/jsons-1.2.0.tar.gz
+  patches:
+  - patches/no-multiprocessing.patch
```
slickepinne, line 8: I don't think it's mandatory, but I'd check that we use a consistent indentation for lists. 
```
  patches:
    - patches/no-multiprocessing.patch
```
This is different in many pyodide packages.

voting_schemes/electionguard/python-wrapper/README.md
```diff
@@ -0,0 +1,21 @@
+# bulletin_board-electionguard
+
+Wrappers used to integrate ElectionGuard python code into the Decidim Bulletin Board system. It consists on three parts:
```
slickepinne, line 3:
```suggestion
Wrappers used to integrate Electionguard python code into the Decidim Bulletin Board system. It consists on three parts:
```

slickepinne (approved):
Great job! And nice to see `pyodide`, again. 😄 I just left some small notes.

Dockerfile
```diff
@@ -42,7 +42,7 @@ ADD decidim-bulletin_board-app/Gemfile.lock /code/tmp/Gemfile.lock
 RUN cd /code/tmp && bundle install
 
 # Add local python dependencies
-ADD voting_schemes/election_guard/python-wrapper /code/voting_schemes/election_guard/python-wrapper
+ADD voting_schemes//python-wrapper /code/voting_schemes/electionguard/python-wrapper
```
orlera, line 45: I'm getting this error when deploying, looks like it is related to this
```
Step 23/33 : ADD voting_schemes//python-wrapper /code/voting_schemes/electionguard/python-wrapper
ADD failed: stat /var/lib/docker/tmp/docker-builder003347322/voting_schemes/python-wrapper: no such file or directory

```
```suggestion
ADD voting_schemes/electionguard/python-wrapper /code/voting_schemes/electionguard/python-wrapper
```

voting_schemes/electionguard/js-adapter/src/trustee_wrapper_adapter.js
```diff
@@ -59,16 +59,14 @@ export class TrusteeWrapperAdapter extends WrapperAdapter {
       }
     );
 
-    if (result) {
+    if (result && result[0]) {
       // eslint-disable-next-line camelcase
-      const { message_type, content } = result;
+      const { message_type, content } = result[0];
```
beagleknight, line 64: As an alternative, you could do this:

```js
if(result && result.length > 0) {
  const [{ message_type, content}] = result;
  ...
```

beagleknight (approved):
Renaming fest! Great job ✨  The docker from docker stuff is so magical 😄

decidim-bulletin_board-app/app/services/voting_scheme/electionguard.rb
leio10: Thanks for the feedback, but my idea is to keep using ElectionGuard as the name of the project, even when the Ruby classes or modules will be called Electionguard to prevent the `_`.

PR #143, opened 2021-03-08
Partial tallying

txus:
After receiving a quorum of trustee shares, if there are some missing we have the option to do partial tallying:

* Bulletin Board sends an empty `tally.compensate` message.
* Each trustee sends a `Compensation` in response to that message.
* Once the Bulletin Board gets all the `Compensation` messages it needs to proceed, it will emit the `end_tally` message with the decrypted tally as usual.


voting_schemes/electionguard/python-wrapper/src/bulletin_board/electionguard/common.py
```diff
@@ -126,16 +149,17 @@ def process_message(self, message_type: str, message: Content) -> Content:
         return responses
 
     def is_fresh(self) -> bool:
-        return isinstance(self.step, self.starting_step)
+        return type(self.step) is self.__class__.starting_step
```
leio10, line 152: 💬 Maybe we could make this method more explicitly. I think this should be only in the Trustee wrapper and we could set a variable to `False` after sending the election keys. What do you think?

voting_schemes/electionguard/python-wrapper/src/bulletin_board/electionguard/serializable.py
```diff
@@ -59,27 +60,27 @@ def maybe_base64_to_int(i: Union[str, int]) -> int:
 
 
 def set_serializers():
-    old_set_serializers()
+    # old_set_serializers()
```
leio10, line 63: ⚠️ should we keep this? if not, we should delete this commented code and the assignment of this function

voting_schemes/electionguard/python-wrapper/src/bulletin_board/electionguard/serializable.py
```diff
@@ -59,27 +60,27 @@ def maybe_base64_to_int(i: Union[str, int]) -> int:
...
     )
     electionguard.serializable.set_serializer(
-        lambda i, **_: int_to_maybe_base64(i), int
+        lambda i, **_: int_to_maybe_base64(i), int  # type: ignore
     )
 
 
 def set_deserializers():
-    old_set_serializers()
+    # old_set_serializers()
```
leio10, line 76: ⚠️ same as before. Also, the method name is wrong, so probably it was not doing anything 🙈

voting_schemes/electionguard/python-wrapper/src/bulletin_board/electionguard/bulletin_board.py
```diff
@@ -1,46 +1,274 @@
...
-    serialize,
-    deserialize,
 )
-from .dummy_scheduler import DummyScheduler
+from .utils import InvalidBallot, deserialize, serialize
+
+TallyResults = Dict[CONTEST_ID, Dict[SELECTION_ID, int]]
+
+
+class TallyDecryptor:
```
leio10, line 38: 💬 This class is pretty big, maybe we could move it to a separate file?

leio10 (approved):
Awesome work!! 💪 👏

slickepinne (approved):
Really nice. I just left some minor notes, and I'm curious about the decryption when some guardians are missing. Would be amazing, if we can check it together when there's time. 😊

voting_schemes/electionguard/python-wrapper/src/bulletin_board/electionguard/serializable.py
txus: Kept it!

voting_schemes/electionguard/python-wrapper/src/bulletin_board/electionguard/bulletin_board.py
txus: Extracted~

voting_schemes/electionguard/python-wrapper/src/bulletin_board/electionguard/serializable.py
txus: Kept it~

.code-workspace
```diff
@@ -0,0 +1,32 @@
+{
+    "folders": [
+        {
+            "name": "Bulletin Board",
+            "path": "decidim-bulletin_board-app"
```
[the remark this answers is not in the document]
txus: Cheers!

voting_schemes/electionguard/python-wrapper/Makefile
```diff
@@ -18,13 +18,16 @@ install-apt:
 install-deps:
 	python3 -m pip install --upgrade pip
 	pip install pipenv
-	cd ../electionguard-python && make environment build
-	pipenv install --dev
-	python3 -m pip install --user --upgrade twine
+	PIPENV_VENV_IN_PROJECT=True pipenv install --dev
+	python -m pip install --user --upgrade twine
```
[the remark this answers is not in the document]
txus: Good one. cheers!

.code-workspace
```diff
@@ -0,0 +1,32 @@
...
+        },
+        {
+            "name": "Bulletin Board Gem",
+            "path": "decidim-bulletin_board-ruby"
```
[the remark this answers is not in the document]
txus: Cheers!

voting_schemes/electionguard/python-wrapper/src/bulletin_board/electionguard/common.py
txus: I'm not sure about making it concrete to the Trustee. I think this is a general question you can ask about any wrapper --is it fresh or has it been restored from some backup?

PR #145, opened 2021-03-09
Move bulk vote generation to background job

orlera:
Moves the bulk votes generation process to a background job in order to avoid long HTTP requests when generating loads of votes.

To give the user a clue when the job has finished, we show the number of votes currently generated at every page load.

[](https://user-images.githubusercontent.com/5033945/110443193-f1dd8b80-80bb-11eb-93fb-e2fa3b5be5d9.mov)


leio10 (approved):
:+1:

slickepinne (commented):
💪

slickepinne (approved):

PR #146, opened 2021-03-09
Generate bulk votes with Electionguard

orlera:
Adds the capability to bulk-generate votes with an EG election.

It:
 - splits the voting scheme service to match the `bulletin_board` and the `voter` adapter, both for the dummy and eg schemes;
 - implements `voter` adapter for both schemas;
 - uses the voter adapter from the bulk generation job inferring the voting scheme from the election.

PS: had to [restrict the test pattern for Jest](https://github.com/decidim/decidim-bulletin-board/pull/146/files#diff-4eda11387cb48905232897cd437c102036ecaa3569a9ffa389e917e9c0bb5de6R144) because in the CI [it was executing tests from dependencies](https://github.com/decidim/decidim-bulletin-board/runs/2078830239?check_suite_focus=true). Not sure if that's related with any of my changes?


decidim-bulletin_board-app/app/jobs/sandbox/generate_votes_job.rb
```diff
@@ -4,21 +4,24 @@ module Sandbox
...
       %w(create_election end_key_ceremony).each do |message_type|
         log_entry = log_entry_for(message_type)
         voter_adapter.process_message(log_entry.message_identifier, log_entry.decoded_data)
       end
 
-      client = Decidim::BulletinBoard::FileClient.new(file_path, OpenStruct.new(client_settings))
       number_of_votes.times do
         encrypted_ballot = voter_adapter.encrypt(random_plaintext_ballot)
-        client.cast_vote(election_id, SecureRandom.hex, encrypted_ballot)
+        client.cast_vote(election_id, voter_id, encrypted_ballot)
```
leio10: Maybe we could leave it as it is now, but I think we will have problems with this (now or in the future), as the server should accept several votes from the same voter_id until they are processed.

leio10 (approved):
Great job! 💟 💎

beagleknight (approved):
Nice 👍

PR #148, opened 2021-03-10
Remove unused cors settings

leio10:
It doesn't make sense to have a setting for allowed CORS domains. In the future we could add all the authorities domains, but at the moment this PR sets that value to `*`.


orlera (approved):

agustibr (changes requested):
👍

We should update the documentation, deleting the `CORS_ORIGIN_ALLOWED` environment variable

https://github.com/decidim/decidim-bulletin-board/blob/d149128bed5aed5a0d38523bca3f073ac196c827/docs/modules/configure/pages/environment_variables.adoc#L9-L10

beagleknight (approved):

PR #149, opened 2021-03-10
Use last ElectionGuard version

leio10:
This PR updates the python wrappers to use the current implementation of EG, to be able to use an election verifier.


beagleknight (approved):
Great 👍

slickepinne (approved):
🎉

PR #150, opened 2021-03-12
Return auditable ballot for ElectionGuard voting scheme

leio10:
This PR makes the voter wrapper to return the encrypted ballot and the auditable ballot to be used by the voting booth.

It also removes the `numpy` dependency from ElectionGuard, as it's only used in one test, but it requires a download of 7mb and makes the voting process slower.


orlera (approved):
📝

beagleknight (approved):
Nice! 🤩

PR #151, opened 2021-03-15
Move javascript assets to gems

beagleknight:
Yet another refactor PR moving files around 😄 . Let me explain it a bit:

We have created a `bulletin_board` folder that contains the `server` (aka `app`), `js-client` and `ruby-client`. The assets that were inside the `server` now are included in the `js-client` and compiled inside the `ruby-client`. In other words, you just need to include the gem in `ruby-client` to work with the bulletin board.

For the `voting_schemes` folder we have added a `ruby-adapter` folder to each voting scheme. These new gems are already used in the bulletin board `server` application and eventually will be used in the `decidim` application as well. At the moment they only include the JS libraries to be used within the asset pipeline.
In a future PR these new gems will also include some ruby code that we will take outside the bulletin board `server` application.

Extra: We don't need to manually copy files anymore! 🎉 
Extra2: In a following up I will create a `Makefile` to handle asset generation and release 👍


orlera (approved):
:hurtrealbad:

bulletin_board/ruby-client/decidim-bulletin_board.gemspec
```diff
@@ -10,7 +10,7 @@ Gem::Specification.new do |s|
 
   s.summary = ""
   s.description = ""
-  s.homepage = "https://github.com"
+  s.homepage = "https://github.com/decidim/decidim-bulletin-board"
```
slickepinne, line 13: 😅 Maybe a description and summary would be nice to have as well.

slickepinne (approved):
This is huge!! 🎉

leio10 (approved):
Well done! ☀️☀️☀️

PR #152, opened 2021-03-15
Improve trustee unique id

slickepinne:
This PR adds the `authority.unique_id` to the `trustee.unique_id` and stores this value in the database. The sandbox and `messages` use the `trustee.slug` which is the `unique_id` without the `authority` part.


beagleknight (approved):
Great job!! ✨ One step forward for consistency 💪

orlera (approved):
👏

decidim-bulletin_board-app/app/models/trustee.rb
```diff
@@ -4,6 +4,14 @@ class Trustee < Client
   has_many :election_trustees
   has_many :elections, through: :election_trustees
 
+  def slug
+    unique_id.split(".").last
```
leio10, line 8: What would happen if the slug given by the authority already has a dot? Maybe you could use `unique_id.split(".", 2)` to be sure that this doesn't affect this (I assume that the `authority_slug` doesn't have dots inside)

leio10 (approved):
Great (and tough) work! 👏 💪

decidim-bulletin_board-app/app/models/trustee.rb
slickepinne: Hmm, yeah. It shouldn't happen, but since we might add the `organization_name` to the `trustee.slug`, there could be a dot in-between. I'll update it. 👍

PR #153, opened 2021-03-16
Add the election verifier

beagleknight:
This is a rough implementation of the election verifier. At the moment it runs 2 verifications:
- A common verification that checks the chained hashes (not depending on the voting scheme).
- A custom verification for `electionguard` that uses a third-party tool tu verify the Electionguard election.

This is a work in progress and it is not meant to be final but I prefer adding features step by step. At the moment the `bin/verify` program can only verify `electionguard` elections. In a future PR we will be able to verify `dummy` ones as well.

The next step is to create another rough implementation for the audit ballot verifier.

If you want to test this locally in your machine you can install `gradle` (https://gradle.org/install/), run `gitsubmodule init` to initialize the `electionguard-java` submodule and run `make test_verifier`.


leio10 (approved):
Awesome!! 👏 👏

agustibr (approved):
😃 Nice!

PR #155, opened 2021-03-17
Add Makefile to install dependencies, build and release all projects

beagleknight:
In this PR my goal was to write a `Makefile` as an entrypoint of many scripts that should be executed to start working in this project. Then I decided to remove the artifacts (javascript assets, python packages, etc.) that were included and that caused many problems with the build and the release process.

About the `Makefile` if you run `make` you will get some help. At the moment if you start from scratch just run all the commands in order. If you already have this repository you can use `make build` to create the missing artifacts or whenever you need to rebuild something.
The `make test` is intended to be used by the CI, but you can run it to run all the tests. Finally, the `make release` will be used for releasing purposes (more below).

Related to the build process since the artifacts are no longer in the repository I changed how the "server workflow" is executed since now I need to build all the artifacts before running all the tests. The workflows for each gem have not been updated yet.

Finally, for the release process we need to discuss two different topics:
- Releasing the three gems is effortless now since you just need to run `VERSION=x.x.x make release` and the script will handle everything for you. I plan to add a manual workflow to release on demand.
- To release the actual server I have been trying to do in Heroku with the new `Dockerfile` but at the moment is not possible. In an upcoming PR I will tackle this problem


leio10 (approved):
✅ very --->  ✅ good --->  ✅ job!!! 
👏👏👏👏👏

agustibr (approved):
👏 Nice work

orlera (approved):
🥡 🥡 🥡

slickepinne (approved):
💙 Great work!!

PR #156, opened 2021-03-18
Update electionguard wrapper and bulletin board server

beagleknight:
This PR includes some updates related to using the latest version of `electionguard-python`. The wrapper has been changed and the server now logs some data to the log entries so the election can be verified.


orlera (approved):
👍🏽

agustibr (approved):
😀

PR #160, opened 2021-03-31
Wait some time before casting the vote while running e2e test suite

beagleknight:
This is another attempt to fix the test flakyness. I tried in #159, but apparently it didn't work. 

The problem is we are receiving "Network errors" while running e2e tests. I suspect that it is something related to the machine running these tests because in my local machine it always works. I'm trying to add some reasonable wait times to make sure it works.


leio10 (approved):

slickepinne (approved):
👍

PR #161, opened 2021-04-05
Add reset_test_database to the bulletin board client

beagleknight:
This adds a new method called `reset_test_database` to the bulletin board client. This method is only available in the `test` environment.

Even if the method is not available the mutation field `resetTestDatabase` is always available to the schema. Calling it in a non-test environment will raise an error.


slickepinne (approved):
👍

leio10 (approved):
👍

PR #162, opened 2021-04-05
Fix scheduler with dummyscheduler

txus:
We accidentally had removed the dummyscheduler, triggering the use of multiprocessing in the voter code which is not allowed.


beagleknight:
I opened #163 to remember to fix the problem with the shares visibility.

beagleknight (approved):

PR #164, opened 2021-04-05
Add conditional ballots in a backwards compatible manner

txus:
* The `encrypt` method on the voter accepts an additional `ballot_style` string parameter that the voter will know from their census / user data. **Backwards compatibility**: if you don't include a ballot style and the election has a single "global" ballot style, that's the one that we'll infer.
* The election can now include `ballot_styles` containing a `contests` key with contest object ids like `['question-1', 'question-2']`, to indicate which ballot styles can vote on which questions.


voting_schemes/electionguard/python-wrapper/src/bulletin_board/electionguard/election.py
```diff
@@ -0,0 +1,123 @@
...
+        ]
+        + synthetic_geopolitical_units,
+        parties=[],
+        candidates=[
+            Candidate.from_json_object(candidate) for candidate in desc["candidates"]
+        ],
+        contests=[],
+        ballot_styles=ballot_styles,
+        name=InternationalizedText.from_json_object(desc["name"]),
+        contact_information=None,  # TODO: add some
```
leio10, line 100: 💬 do we need to add something here or can we delete the TODO?

leio10 (approved):
Awesome work! Using EG objects is much better than the previous approach! 😍 👏

voting_schemes/electionguard/python-wrapper/src/bulletin_board/electionguard/election.py
txus: Yep, removed the todo

PR #166, opened 2021-04-12
Update rails dependency

beagleknight:
This update the `rails` dependency and remove the `mimemagic` dependency. I also changed the name of the npm packages and add them to the `make release` task.


orlera (approved):
**yanked**

PR #167, opened 2021-04-12
Add ballot styles to the voter client

beagleknight:
In this pull request I am using #164 in the frontend part. When voting you can select the ballot style you want to use if any.

I needed to change the sandbox a bit so I added some improvements:
- When creating an election you can add multiple questions.
- When creating an election you can add multiple ballot styles. If you add a question afterwards all ballot styles are updated accordingly.
- When voting you can select the ballot style you want to use.
- When voting you can answer questions using a simple form.

https://user-images.githubusercontent.com/106021/114390619-c9ebb700-9b96-11eb-9b77-1ea05ee3974a.mov

https://user-images.githubusercontent.com/106021/114390999-3c5c9700-9b97-11eb-90c8-fd96232d0d25.mov

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+## Added
+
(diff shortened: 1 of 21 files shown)


orlera (approved):
🗳️

PR #168, opened 2021-04-13
Fix dev containers

leio10:
This PR updates the VS Code dev containers configuration to be able to use the new base Docker image and the Makefile to install all the dependencies and build all the needed artifacts.
It also adds a `replant_serve` target to have an easy way to start a BB server with a clean database with the app seeds.


orlera (approved):
❤️

agustibr (approved):

beagleknight (approved):
Nice ✨

PR #169, opened 2021-04-14
Use the correct main file in the bulletin board js client

beagleknight:
This fixes the bulletin board javascript client so it can be used in `decidim` with `Webpacker`.


orlera (approved):
🔨

agustibr (approved):

PR #170, opened 2021-04-14
In person vote

leio10:
This PR allows authorities to register in person votes, to be able to ignore duplicated votes for the same voter. To do this, the authority should register the list of polling stations during the election creation, and then assign each in person vote to a polling station.

This PR add the server support for this message, refactors the tally to avoid counting duplicated votes and modify the BB client to allow the authorities to use this feature.

A future PR will apply this feature to the sandbox, along other improvements on the election creation and the voting process.


bulletin_board/server/Gemfile
```diff
@@ -23,7 +23,7 @@ gem "turbolinks", "~> 5"
 # Build JSON APIs with ease. Read more: https://github.com/rails/jbuilder
 gem "jbuilder", "~> 2.7"
 # Use Redis adapter to run Action Cable in production
-gem "redis", "~> 4.0"
+# gem "redis", "~> 4.0"
```
orlera, line 26: Why are we removing `redis`? 👓

orlera (approved):
👏🏽  that's a big step forward! 🤩

bulletin_board/server/Gemfile
leio10: Nope, but I think we are not using that gem anymore, as we don't use Action Cable. Redis will be used when using Sidekiq for ActiveJobs, but currently is not configured.

agustibr (approved):
🙌 Great!

bulletin_board/server/app/services/valid_votes.rb
```diff
@@ -0,0 +1,21 @@
...
+    @election = election
+  end
+
+  def each
+    counted_votes = ElectionInPersonVotes.for(election).pluck(:author_unique_id).to_set
+    ElectionCastedVotes.for(election).find_each(order: :desc) do |vote|
+      next if counted_votes.include?(vote)
+
+      yield vote
+      counted_votes << vote
```
beagleknight, line 14: As a minor refactor you could do something like:
```ruby
ElectionCastedVotes.for(election).where.not(id: counted_votes).find_each(order: :desc) do |vote|
  # if is no longer needed
end
```

beagleknight (approved):
LGTM! 👍 Just added a small suggestion.

bulletin_board/server/app/services/valid_votes.rb
leio10: That's a very interesting suggestion, for several reasons:
* There is a silly bug that I didn't see: the `counted_votes` set should have voter ids, not votes. I didn't see it because I forgot to add the test for that class. 😞 I'll add it now.
* Regarding including the list of counted votes in the query, I'm not sure because of the possible size of the set. I've made some tests to be sure that the app can have a set with 1M hashes of 32 bytes in memory and it works ok. But I tried to send that hashes in a query to the database and it works (!) but it's very very slow.
* But thanks to your idea I've found an intermediate solution to avoid loading votes from voters that have voted in person that I its much better. We can build a compound query to filter those hashes directly on the database.
```ruby
ElectionCastedVotes.for(election).where.not(
  author_id: ElectionInPersonVotes.for(election).select(:author_unique_id)
)
```
If I add `to_sql` to that query I got this, that it's exactly what I want: 
```sql
SELECT "log_entries".* FROM "log_entries" 
WHERE "log_entries"."election_id" = 23 AND
      "log_entries"."message_type" = 'vote' AND
      "log_entries"."message_subtype" = 'cast' AND
      "log_entries"."author_id" NOT IN (
            SELECT "log_entries"."author_unique_id" FROM "log_entries" 
            WHERE "log_entries"."election_id" = 23 AND 
                  "log_entries"."message_type" = 'vote' AND 
                  "log_entries"."message_subtype" = 'in_person'
      )
```
* Even with that, I cannot remove the `if`, as the idea is to avoid processing several votes for the same `voter_id`. I'm not sure if I can remove them directly in the database, because the use of `find_each` to load records per batches limits the SQL query that can be used.

beagleknight: That's really good news! Good job 😄

beagleknight (approved):
Nice! 👍

PR #171, opened 2021-04-14
Audit vote using verifier

beagleknight:
In this pull requests my goal was to add the ballot verifier for the `electionguard` scheme. At the moment the ballot verifier also contains a common verifier that checks if the content hash is correct.

You can test it like this (assuming that you have the bulletin board server started at port 8000):
```
ELECTION_UNIQUE_ID=<ELECTION_UNIQUE_ID> make test_verify_ballot
```

The previous command will test the ballot included in the `verifier/test/fixtures/electionguard/ballot-ok.txt`. Before running the command you should audit a ballot and move the file to this folder and then run the command using the correct election unique id.

Let me explain a bit how the whole process works:
- First it parses the ballot file as a JSON
- Then it checks the encrypted data hash (common verifier)
- Finally it runs a python script that encrypts the vote using the auditable data. The data contains some values that make the encryption deterministic, so we can just test if we get the same result.

When I was doing this, I decided to migrate the code to `ESModules` instead of `CommonJS` modules. It is well-supported since Node 12, so there is no reason to use `CommonJS` modules anymore.


voting_schemes/electionguard/verifier/src/services/bulletin-board-client.service.js
```diff
@@ -0,0 +1,62 @@
...
+import fetch from "node-fetch";
+import InMemoryCache from "apollo-cache-inmemory";
+import gql from "graphql-tag";
+
+/**
+ * This class includes some methods to perform requests to the Bulletin Board
+ * server GraphQL API.
+ * @class
+ */
+export class BulletinBoardClient {
```
orlera, line 12: At some point we could think of using the BB npm package, once it'll be ready to be used with node 🚀

orlera (approved):
Yay! 🗳️ ✅

leio10 (approved):
Good work! 👏 👏

PR #173, opened 2021-04-16
Enhance ballot to make the audit agnostic from voting scheme

beagleknight:
This adds a new verifier for the ballot: the questions and answers verifier. You can see a quick demo below.

When a voter audits a ballot they will download a file that contains both the plain vote and the election unique id. With this data, the verifier will print the questions and answers and mark the voter answers sot hey can verify that the answer was correct. 

Since I am adding the election unique id to the data the `ELECTION_UNIQUE_ID` environment variable is no longer needed (see #171)

https://user-images.githubusercontent.com/106021/115016038-d5035780-9eb4-11eb-8a2c-415cc00e11bf.mov


leio10 (approved):
Great job! The verifier output looks great! 🤩 🤩 

Just a small question: As the plain vote is now included by the voter wrapper, shouldn't we modify the returned data from the dummy and EG encrypt methods?

beagleknight:
> Great job! The verifier output looks great! 🤩 🤩
> 
> Just a small question: As the plain vote is now included by the voter wrapper, shouldn't we modify the returned data from the dummy and EG encrypt methods?

@leio10 sure we can do that since it is no longer necessary! I would consider it a refactor though, since it doesn't harm at the moment 😄

orlera (approved):

PR #174, opened 2021-04-19
Generate verifiable election before publishing results

### LINKED ISSUE #163: Hide trustee share data when the results are not published
We should hide the trustee share data from the election log entry data until the election results are published because any user could compute the result with that data.

After discussing this with @leio10 and @txus we reach the conclusion that this can be solved in two different ways:
- Hide the data when the tally has ended
- Authenticate the trustee using some kind of signature verification

beagleknight:
This restores the message visibility that we commented temporary in #162. From now on, a trustee will be able to see this hidden data because they will be authenticated using a simple mechanism: when they use the GraphQL client they will send a signed message that the server will verify.

I also moved the verifiable election file to the end of the tally process. The authority will be able to get this url and verify the election before publishing the results.

The GraphQL fields visibility is checked using `guards`. I added the `graphql-guard` gem and implemented a basic policy.

Finally, I upgraded `cypress` to version 7 because it didn't include breaking changes for us, and it may make our tests less flaky 🤞 

Related issues:

- Closes #163


orlera (approved):
Nice! 👏🏽

bulletin_board/server/Gemfile.lock
```diff
@@ -360,6 +362,7 @@ DEPENDENCIES
   graphiql-rails
   graphlient (~> 0.4.0)
   graphql (~> 1.9)
+  graphql-guard (~> 2.0)
```
leio10, line 365: This gem is awesome!! 🤩

bulletin_board/server/app/graphql/policies/visibility_for_all.rb
```diff
@@ -22,8 +22,12 @@ def self.trustee?(context)
       trustee = Client.where(unique_id: trustee_unique_id).first
       return false if trustee.blank?
 
-      decoded_api_key = JWT.decode(context[:api_key], trustee.public_key_rsa, true, verify_iat: true, algorithm: "RS256").first.with_indifferent_access
-      decoded_api_key[:trustee_unique_id] == trustee_unique_id
+      begin
+        decoded_api_key = JWT.decode(context[:api_key], trustee.public_key_rsa, true, verify_iat: true, verify_expiration: true, algorithm: "RS256").first.with_indifferent_access
```
leio10: mm, there is no need to verify the `iat`, right?

leio10 (approved):
Well done!! 👏 🔝

bulletin_board/server/app/graphql/policies/visibility_for_all.rb
beagleknight: Dunno 😅 . Do you think we can remove this check?

PR #175, opened 2021-04-22
Expose verifiable results in get results GraphQL query

orlera:
The ruby client now also asks for the election's verifiable data (the hash and the file's path) with the `get_election_results` query.
This changes slightly the structure of the response, that will need to be adapted in Decidim.

The sandbox has been updated to show the new data retrieved.

image


leio10 (approved):
Great! 👏 💎

beagleknight (approved):
Nice! 👏

PR #176, opened 2021-04-22
Return data for election log

slickepinne:
Now, when querying for `GetElectionLogEntries`, you get the `chainedHash` as well. Also, the `MessageParser` is exported. With that, you can now access the `iat` from Decidim without exposing the whole `decodedData`.

The `chainedHash` will be part of the `Election Log` on Decidim.


leio10 (approved):
Good job! 👍

orlera (approved):
🎸

beagleknight (approved):

PR #178, opened 2021-04-26
Use the correct docker image as base image

beagleknight:
While I was deploying this to `staging` I detected some problems related to the docker base image that includes everything (ruby, node, python and electionguard).

I decided to change its name schema (it's horrible, I know 😅 ) to include all the versions to avoid problems in the future. Previously it was just using the SHA of electionguard commit but if we changed something like the python version the name didn't reflect the change.

Also, I realized that we were including some config data like `.venv` in the docker image mistakenly. This caused some problems while building the docker image in my machine.


orlera (approved):

agustibr (approved):
👍

leio10 (approved):

PR #179, opened 2021-04-28
Add the verify binary

beagleknight:
This adds a simple binary that will be used to verify either an election or a ballot. 

```sh
bin/verify <my_election_or_ballot_file_path>
```


leio10 (approved):
❤️

orlera (approved):
🚀

PR #197, opened 2021-08-17
Add alias for unique_id

slickepinne:
We renamed the `unique_id` method to `slug`. Therefore, we got an error when trying to retrieve the `id` of the `client`. This PR adds an `alias` for the `slug` method.

query with alias method


agustibr (approved):
🎏 👍

leio10 (approved):
👏 Awesome!

PR #227, opened 2022-02-10
Clarify on documentation the necessary checks when setting up an election

paarals:
Before setting up the election, it is a good idea for the administrator to know which checks are expected to be performed


andreslucena (approved):
👍🏽

PR #253, opened 2022-03-26
Audit improvements

txus:
- [x] Verify trustee keys during ceremony
- [x] Use JSON serialization instead of Marshal
- [x] Ensure the chained hash will use election id only on the first message
- [x] Document how to use cloud storage

### DIFF
--- a/bulletin_board/server/config/environments/production.rb
+++ b/bulletin_board/server/config/environments/production.rb
@@ -38,7 +38,7 @@ Rails.application.configure do
   # config.action_dispatch.x_sendfile_header = 'X-Accel-Redirect' # for NGINX
 
   # Store uploaded files on the local file system (see config/storage.yml for options).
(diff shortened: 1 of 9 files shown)


bulletin_board/server/config/environments/production.rb
```diff
@@ -38,7 +38,7 @@
   # config.action_dispatch.x_sendfile_header = 'X-Accel-Redirect' # for NGINX
 
   # Store uploaded files on the local file system (see config/storage.yml for options).
-  config.active_storage.service = :local
+  config.active_storage.service = :amazon
```
andreslucena, line 41: This is against Decidim Social Contract. You can't force implementers to use a privative external service. Please change it and do it an optional env var

andreslucena: Reported at #261

PR #265, opened 2022-03-31
Add .rbenv-vars to the .gitignore

ahukkanen:
Not to commit local env vars when using [rbenv-vars](https://github.com/rbenv/rbenv-vars), add `.rbenv-vars` to the `.gitignore`.


andreslucena (approved):
I agree, we also have them in [decidim's gitignore](https://github.com/decidim/decidim/blob/6144323a75f943756b2c79a49b35e32703313405/.gitignore#L22) so just for consistency LGTM

PR #284, opened 2022-05-08
Fix GitHub build

txus:


andreslucena:
The unrecognized cop error is related to the rubocop update at Decidim repository (see https://github.com/decidim/decidim/pull/9201), so updating to the same version should solve that.

andreslucena:
Closing this one in favour of #286 
Thanks for this PR @txus!

PR #286, opened 2022-06-29
Fix GitHub build

andreslucena:
This takes over the work from @txus in #284, adding the necessary Rubocop fixes


bulletin_board/server/app/commands/end_vote.rb
```diff
@@ -24,13 +24,13 @@ def initialize(authority, message_id, signed_data)
...
+      valid_client?(authority.authority? && election.authority == authority) &&
+      valid_author?(message_identifier.from_authority?)
 
     election.with_lock do
-      return broadcast(:invalid, error) unless
-        valid_step?(election.vote?) &&
-        valid_client?(authority.authority? && election.authority == authority) &&
-        valid_author?(message_identifier.from_authority?) &&
-        process_message
+      return broadcast(:invalid, error) unless process_message # rubocop:disable Rails/TransactionExitStatement
```
andreslucena, line 33: Note to reviewer: I had to leave this exit in the lock, as this method (`process_message`) is actually leaving the `election` dirty (without saving the model) 

https://github.com/decidim/decidim-bulletin-board/blob/43b5d84630fb974c4e7a8d5cbfd93eb9b7b22dac/bulletin_board/server/app/commands/concerns/log_entry_command.rb

This generated an exception: 

>       RuntimeError:
>         Locking a record with unpersisted changes is not supported. Use `save` to persist the changes, or `reload` to discard them explicitly.

andreslucena:
@ahukkanen this is ready to be reviewed, can you give it an eye when you have the time? Thanks

ahukkanen:
Since the Rubocop rules are inherited from decidim/decidim and we've upgraded Decidim already to Ruby 3.1, I think we have to wait until we can upgrade this repo also to Ruby 3.1.

The main problem right now is explained at #288, i.e. compatibility with OpenSSL 3.0 is still pending for one of the dependencies. It's actually resolved already but we are still waiting for a release for that gem.

ahukkanen (approved):

PR #266, opened 2022-03-31
Update getting started guide to the README

ahukkanen:
This adds some basic instructions to get started with this repository.

Even when using devcontainer, I think this is useful to better understand the different parts and what is needed for the build process. I personally also prefer to work locally and not inside a devcontainer (especially when doing the Ruby 3 update).

PR #291, opened 2022-08-26
Define local python version for pyenv

ahukkanen:
As per review of #266, there was a comment that pyenv does not automatically pick up the python version:
https://github.com/decidim/decidim-bulletin-board/pull/266/files#r902628186

This fixes the issue by defining a local penv version.


andreslucena (approved):
Great! 👍🏽

PR #229, opened 2022-02-10
Explain on documentation that you can only export one time the access codes

paarals:
saying that you can only export access code 1 time

PR #295, opened 2022-11-11
Explain on documentation that you can only export one time the access codes

ahukkanen:
Let's take over #229 since it's stale and lives in a forked repo.

This improves the docs regarding exporting the access codes.


andreslucena (approved):
👍🏽

PR #294, opened 2022-09-12
Fix issue with OpenSSL 3.0

ahukkanen:
There is still a compatibility issue in the Bulletin Board with OpenSSL 3.0.

The comment is stating that after jwt/ruby-jwt#375 is merged and released, the extra stuff in the import code is no longer needed which seems to make sense looking at the changes in that PR. These changes became part of [jwt-2.2.3](https://github.com/jwt/ruby-jwt/releases/tag/v2.2.3) (2021-04-19), so these should no longer be needed.

When running the current code under OpenSSL 3.0, you will see the following exception:
```
     @bulletin_board ||= Decidim::BulletinBoard::Client.new
     
     OpenSSL::PKey::PKeyError:
       rsa#set_key= is incompatible with OpenSSL 3.0
     # /home/ahukkanen/.rbenv/versions/3.1.1/lib/ruby/gems/3.1.0/gems/decidim-bulletin_board-0.24.0/lib/decidim/bulletin_board/jwk_utils.rb:19:in `set_key'
     # /home/ahukkanen/.rbenv/versions/3.1.1/lib/ruby/gems/3.1.0/gems/decidim-bulletin_board-0.24.0/lib/decidim/bulletin_board/jwk_utils.rb:19:in `import_private_key'
     # /home/ahukkanen/.rbenv/versions/3.1.1/lib/ruby/gems/3.1.0/gems/decidim-bulletin_board-0.24.0/lib/decidim/bulletin_board/settings.rb:13:in `initialize'
     # /home/ahukkanen/.rbenv/versions/3.1.1/lib/ruby/gems/3.1.0/gems/decidim-bulletin_board-0.24.0/lib/decidim/bulletin_board/client.rb:26:in `new'
     # /home/ahukkanen/.rbenv/versions/3.1.1/lib/ruby/gems/3.1.0/gems/decidim-bulletin_board-0.24.0/lib/decidim/bulletin_board/client.rb:26:in `initialize'
```

Therefore, we should remove this code as stated in the comment and let `jwt` itself handle this part.

PR #296, opened 2022-11-11
Bump version to 0.24.1

ahukkanen:
In order to release a new patch version of the bulletin board packages/gems and the voting schemes, I am bumping the version number to 0.24.1 and adding a CHANGELOG entry for this.

The main reason for the new release is the OpenSSL 3.0 fix implemented at #294.

We need this release in order to finish decidim/decidim#9766.


ahukkanen:
@andreslucena I need one approving review to be able to merge this, so I'd appreciate if you could check it.

I am only bumping the version to 0.24.1 (through `VERSION="0.24.1" make bump_versions`) in order to release the new version with the important fix.

CHANGELOG.md
```diff
@@ -7,6 +7,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+## [0.24.1] - 2022-11-11
+
+## Fixed
+
+- Fixed issue with bulletin board Ruby client OpenSSL 3.0 compatibility
```
andreslucena, line 14: This was manually added? Should we add the other changes too? 

https://github.com/decidim/decidim-bulletin-board/compare/v0.24.0...develop

andreslucena (changes requested):
I have only one doubt regarding the CHANGELOG and it'd be ready to go for me.

CHANGELOG.md
ahukkanen: Yes, this is manually created as we don't have automation in this repo.

Aren't they mostly documentation / local configuration changes?

I think they are not relevant in the changelog, I would categorize them in the Decidim main repo as `internal`.

Also looking at the previous entries in the changelog, I don't think it includes everything.

andreslucena: Ok, maybe we'd need to standardize the labels/changelog generation/etc to be consistent with `decidim/decidim`. But at the moment as this project isn't used in production I think that what we have is good then.

andreslucena (approved):
👍🏽

PR #301, opened 2023-06-01
Change the Elections' setup minimum hours before start setting to 1 hour

andreslucena:
This is the other part of https://github.com/decidim/decidim/pull/10900. See that for more details and the explanation, but basically: working with Elections is painful because you need to wait a lot of time between tryouts. This PR fixes it by changing the default parameter to 1 hour.


microstudi:
superseded by #310

PR #304, opened 2023-09-21
Bump electionguard from 1.2.1 to 1.2.3 and fix build

microstudi:
This project uses a lot of external libraries, many of them on very active development. And, in some places, relays on instructions of the type "install the latest version" or "bigger than that version". As the libraries advance, tend to deprecate old dependencies and this "last versions" break on compilation.

In this case, ElectionGuard 1.2.1 has a dependency with `numpy` >1.8.2, which in the latest version is not compatible with python 3.8 (used in this project). 
Ideally we should upgrade ElectionGuard to the latest version (goes for the 1.4 now) but I suspect is going to be a major headache.
For the moment, we can upgrade to ElectionGuard 1.2.3 which is a minor version and should keep any compatibility with the project.

I found some problems while building the project, also related with other libraries like pyodide 0.16.1 (which is heavily outdated). It uses a Dockerfile to build and intermediate container that uses `wget` with old certificates and it does not download properly a dependency (libgmp). 

This PR solves all those things while trying not to touch any of the hacks/patches introduced in many places. About this, we should have a discussion on how to document the reason of all these patches because it makes very difficult to upgrade the project to more modern versions (ElectionGuard is patched, Pyodide is patched but nowhere is explained why). Documenting these should make easier to upgrade to more modern versions of the libraries as we would know if we still need those patches for instance.

If you apply this PR, 2 docker images should be built and pushed to the docker registry, these are:
- decidim/ruby-node-python-electionguard:ruby-3.1.1-node-16-python-3.8.11-electionguard-1.2.3
- decidim/pyodide-electionguard:pyodide-0.16.1-electionguard-1.2.3

It can be done by executing in the root the following commands (assuming you have permission to push to the docker registry):

```
make electionguard_docker_pyodide
make electionguard_docker_base
```
Note that probably tests on these PR won't work until those two commands are executed on someone's machine with permissions to push to the docker registry

PR #305, opened 2023-09-21
Fix error while building the pyodide-electionguard docker image

andreslucena:
After #304 was merged, I thought we could build the electionguard docker images, but I found the same error with the certificates for gmplib.org 

In this case, it wasn't enough with updating `wget`, but updating `libgnutls28-dev` made the trick. I found this package at [gmplib.org mailing list](https://gmplib.org/list-archives/gmp-discuss/2020-June/006492.html) 

For easing up these tests, I separated the build and push targets in the Makefile

## Testing

Both of these commands should run without problems:
```shell
make build_electionguard_docker_base
make build_electionguard_docker_pyodide
```


microstudi (commented):
Lgtm! Please merge!

microstudi (approved):

PR #306, opened 2023-09-22
Fix python wrapper after upgrading electionguard to 1.2.3

microstudi:
After updating to ElectionGuard 1.2.3 the libray `electionguard.types`  has been renamed in the python package. This PR adapts this to the pyodide wrapper.

This PR also organizes a bit better the different Dockerfiles (I guess there are 3 because of heroku, but only Dockerfile is needed in order to work).


docker-compose.yml
andreslucena, line 1: When I run 
``` 
docker-compose up
``` 

I have the following error message: 

```
rake aborted!                                                                            
ActiveRecord::ConnectionNotEstablished: could not connect to server: Connection refused  
  Is the server running on host "localhost" (127.0.0.1) and accepting    
  TCP/IP connections on port 5432?                                                       
could not connect to server: Cannot assign requested address
  Is the server running on host "localhost" (::1) and accepting
  TCP/IP connections on port 5432?
```


Can you provide the documentation/steps for this to work?

docs/modules/develop/pages/manual/operations/deployment.adoc
```diff
@@ -28,13 +33,23 @@ We'll need to set up an array of environment variables in order for the sever to
...
+ ** For instance, if using a compatible AWS provider you should define:
+   *** `ACTIVE_STORAGE_SERVICE=s3`
+   *** `AWS_ACCESS_KEY_ID=********`
+   *** `AWS_SECRET_ACCESS_KEY=********`
+   *** `AWS_REGION=eu-west-1`
+   *** `AWS_BUCKET_NAME=bucket-name`
+   *** If using a provider that is not Amazon also define `AWS_ENDPOINT` with the endpoint of the provider.
+ ** You can also easily use Azure or GCS providers (see the previous mentioned `storage.yml` file for the ENV vars), but in this case add the line that applies in your case to the file `bulletin_board/server/Gemfile`:
+   *** `gem "azure-storage-blob", require: false`
+   *** `gem "google-cloud-storage", "~> 1.11", require: false`
```
andreslucena, line 52: Can we add it commented out in the Gemfile? Also, can you add the `azure-storage-blob` version too?

Dockerfile.web
andreslucena, line 1: Why are we changing the Dockerfile.web to Dockerfile?

These kind of changes make more difficult to work with the git history of the project. For instance, in the GH Files Tab, it doesn't recognize that's the same file, so I'd need to make the diff manually to see if there's actually any change apart of the rename.

So, is there any reason why doing this change?

andreslucena (changes requested):
I have some doubts, can you check them out please?

Dockerfile.web
microstudi: Because Dockerfile.web is just not standard, usually if you find a Dockerfile you know is the right one. And I found that all 3 Dockerfiles were exactly the same, just with lines switched in order. To be honest, I think we should have only one, but because there is this love for heroku I left their versions.

docs/modules/develop/pages/manual/operations/deployment.adoc
microstudi: sure, as a note, there's no version in the rails documentation when recomend adding this line

docker-compose.yml
microstudi: Database file was missing, now should be working.

docker-compose.yml
```diff
@@ -0,0 +1,54 @@
+version: "3.3"
+services:
+  db:
+    image: postgres:14
+    environment:
+      - POSTGRES_PASSWORD=postgres
+      - POSTGRES_DB=decidim_bulletin_board
+    volumes:
+      - pg-data:/var/lib/postgresql/data
+  bulletin:
```
andreslucena: I had to double check that this was actually another service.

Also, I prefer to use the `app` service naming as its more standard (it could be `web`, whatever you prefer, as long is not `bulletin` 😄)  
```suggestion
      - pg-data:/var/lib/postgresql/data

  app:
```

docs/modules/develop/pages/manual/operations/deployment.adoc
andreslucena: Ok, if the docs don't have it then I can live with that

docker-compose.yml
andreslucena: Probably `web`  is better as that's the one used in the heroku.yml

andreslucena (approved):
👍🏽

Dockerfile.web
andreslucena: Ok, in a next release we can see about consolidating them on only one Dockerfile. 
Also, we should not offer support to a proprietary hosting service, specially heroku that seems to be in their way out.

PR #302, opened 2023-06-01
Fix the rubocop configuration URL

andreslucena:
With https://github.com/decidim/decidim/pull/10578 we've changed where the rubocop configuration is on the main repository. 

Currently some linting workflows are broken in the `develop` branch: 

https://github.com/decidim/decidim-bulletin-board/actions/runs/5142715203/jobs/9256792760
https://github.com/decidim/decidim-bulletin-board/actions/runs/5142715210/jobs/9256792775

This PR fixes it on this repository.

PR #308, opened 2023-09-28
Update pyodide to 0.17 and fix js adapter

microstudi:
This PR makes the JS client, server and pyodide adapters work again.

- Upgrade pyodide to 0.17: This new version of pyodide changes some transformations between python and javascript and some adaptations are needed-
- update base docker image for pyodide and electionguard 1.2.3
- Update several libraries and make the test happy again
- supersedes #165 
- supersedes #302 
- supersedes #299 

Note:
After this PR is merge I'll update #309 to the next version number, test it on the decidim PR upgrade branch (https://github.com/decidim/decidim/pull/11711). When everything green we can release the final version


bulletin_board/server/package.json
```diff
@@ -37,7 +37,7 @@
     "eslint-plugin-jasmine": "^4.1.1",
     "eslint-plugin-node": "^11.1.0",
     "eslint-plugin-promise": "^4.2.1",
-    "prettier": "2.1.2",
+    "prettier": "3.0.3",
```
andreslucena, line 40: Are you changing this because there was some error in the pipeline related to the old version of prettier? I say because I prefer to separate the linter changes from the other changes, as they bring too much noise to the reviews.

voting_schemes/electionguard/python-wrapper/src/bulletin_board/electionguard/bulletin_board.py
```diff
@@ -14,7 +14,7 @@
 from electionguard.type import GUARDIAN_ID
 from electionguard.utils import get_optional
 
-from .common import Content, Context, ElectionStep, Wrapper
+from .common import Content, Context, ElectionStep, Wrapper, async_wrap
```
andreslucena: Is `async_wrap` being used in this file?

voting_schemes/electionguard/python-wrapper/Pipfile
```diff
@@ -6,6 +6,7 @@ verify_ssl = true
 [dev-packages]
 flake8 = ">=3.8.3"
 pytest = "*"
+pytest-asyncio = "*"
 black = "*"
```
andreslucena, line 10: Shouldn't we pin these versions just to be sure that this doesn't break in an update?

bulletin_board/server/cypress/integration/sandbox/election.page.js
```diff
@@ -109,15 +109,15 @@ export class ElectionPage {
...
+              },
             );
 
           cy.findByText("Start").click();
 
           cy.findByText(name)
             .parent("tr")
             .within(() => {
-              cy.findByText("Backup").click({ timeout: 180_000 });
+              cy.findByText("Backup").click({ timeout: 720_000 });
```
andreslucena, line 120: Is it really necessary quadrupling this timeout or maybe just duplicating could be enough? 
```suggestion
              cy.findByText("Backup").click({ timeout: 360_000 });
```

voting_schemes/electionguard/docker/pyodide-electionguard/packages/libgmp/Makefile
andreslucena, line 1: Just to see if I understood the change, this means that we don't need to change the configure with different flags the libgmp (lines 35 to 39) and we can do it with the vanilla libgmp-dev from the apt repositories?

andreslucena (commented):
I have some questions, but generally is looking good 😄

bulletin_board/server/package.json
microstudi: All changes are for necessity, without this change the linter tests don't pass.

voting_schemes/electionguard/python-wrapper/Pipfile
microstudi: but all the other dependencies aren't blocked to a version. This justs adds async library allowed by pyodide 17

bulletin_board/server/cypress/integration/sandbox/election.page.js
microstudi: No, it's not. Already tested (just look at the previous commit)

voting_schemes/electionguard/docker/pyodide-electionguard/packages/libgmp/Makefile
microstudi: it's the same written diffently, look at the `meta.yaml` file, there are the same flags. This comes from the original PR trying to upgrade to pyodide 17

voting_schemes/electionguard/python-wrapper/src/bulletin_board/electionguard/bulletin_board.py
microstudi: looks like not. This also comes from the original PR. Do yo want me to remove it?

microstudi:
@andreslucena what you say, can we merge this?

bulletin_board/server/package.json
andreslucena: Okey makey!

bulletin_board/server/cypress/integration/sandbox/election.page.js
andreslucena: I think you mean to say "Yes, it is" 😄 
I've missed that commit, it makes sense!!

voting_schemes/electionguard/docker/pyodide-electionguard/packages/libgmp/Makefile
andreslucena: Ok, but I don't fully understand why we're removing this file actually, but if it builds without it then its better for sure

voting_schemes/electionguard/python-wrapper/Pipfile
andreslucena: I'm not familiar with the modern Python tooling. In my time people did `pip freeze > requirements.txt` and that was it. I understand that the `Pipfile` is more familiar to the `Gemfile`, and at least with bundler this would be a code smell.

As I'm not sure if "*" is considered a bad practice in pipenv, I'll let it pass.

voting_schemes/electionguard/python-wrapper/src/bulletin_board/electionguard/bulletin_board.py
andreslucena: Yes, please. As far as I see this is imported but not used, so it seems like its not doing anything. If that's not the case and there's some failing specs after this change, we can reintroduce it.

andreslucena (approved):
👍🏽

